PR #21, opened 2026-03-17
ai: Added OpenAI compatible providers

xpahos:
Implement OpenAI-compatible provider to support GLM, Kimi, Minimax, and OpenAI APIs through a unified client that uses model-based defaults and allows configuration overrides for endpoints and context windows

### DIFF
--- a/src/bin/benchmark_review.rs
+++ b/src/bin/benchmark_review.rs
@@ -16,7 +16,10 @@ use anyhow::{Context, Result};
 use clap::Parser;
 use futures::stream::StreamExt;
 use regex::Regex;
-use sashiko::ai::gemini::{Content, GeminiClient, GenAiClient, GenerateContentRequest, Part};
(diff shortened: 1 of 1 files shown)


xpahos:
> Quick question: have you actually tested it with any of these models? 

Yes, I checked this using GLM. But I decided to check OpenAI as well, and there are slight differences compared to other models (`max_tokens`).

> Can you, please, run at least a small benchmark (100 test cases)?

The free, unlimited models available to me are quite slow. It will take a while, but I’ve started the tests.

Today I’ll try to come up with a good solution for `max_tokens` across different models, and then I’ll update the pull request.

rgushchin:
I think that supporting OpenAI models is really useful. Can you, please, remove the mentioning of all other specific models which you haven't excessively tested (GLM, Kimi, Minimax, etc) and I'm going to merge it.

PR #25, opened 2026-03-17
feat: log AI request/response content per turn

paf0186:
## Summary
- Add per-turn logging of AI requests and responses, gated behind a new `log_turns` setting in `[ai]` (default: false)
- When enabled, logs each outbound request with turn number, message count, role, and content/tool-call preview
- Logs each response with text preview, tool calls, and token usage breakdown (input/output/cached)

Useful for debugging runaway or unexpected reviews without adding noise to normal operation.

## Test plan
- [ ] Run a review with `log_turns = false` (default) and verify no per-turn lines appear
- [x] Run with `log_turns = true` and verify turn logging appears at info level
- [x] Confirm log lines are truncated and don't produce excessively long output

### DIFF
--- a/src/reviewer.rs
+++ b/src/reviewer.rs
@@ -1425,6 +1425,7 @@ async fn run_review_tool(
             let mut ai_started = false;
             let mut total_tokens_used: usize = 0;
             let mut total_output_tokens_used: usize = 0;
+            let mut turn_count = 0u32;
 
             loop {
                 let line_result = match timeout_at(deadline, lines.next_line()).await {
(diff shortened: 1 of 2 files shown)


Rongronggg9:
I found this PR useful when I was messing around with self-hosted LLMs with the openai-compatible backend. It would be more useful if there were an option to specify the length limit so that debugging with invalid JSON responses would be easier, as Sashiko heavily depends on the API returning valid JSON responses.

paf0186:
@Rongronggg9 that's a good addition; if you wanted to update + rebase the PR, be my guest

PR #43, opened 2026-03-24
feat: add claude-cli provider for Claude Code subscription auth

paf0186:
## Summary

New AI provider that shells out to `claude --print` instead of calling api.anthropic.com directly. This lets sashiko use Claude Code's subscription authentication, avoiding the need for separate API credits.

**Why this is interesting:** This is a CLI-based provider — it spawns `claude --print` as a subprocess, writes the prompt to stdin, and parses the JSON response from stdout. This means anyone with a Claude Code subscription can run sashiko reviews without API keys or credit management. The tradeoff is no streaming and slightly higher latency per turn, but for batch review workloads this is fine.

### Implementation details

- `src/ai/claude_cli.rs`: Full `AiProvider` trait implementation
  - Spawns `claude --print --output-format json --verbose --no-session-persistence`
  - 10-minute timeout per CLI call (prevents hung processes)
  - Robust JSON parsing: handles tool calls, multi-object responses, nested JSON
  - Token usage extracted from Claude's verbose output
  - Proper stdin flush + close before waiting for output
- `src/ai/mod.rs`: Registers `"claude-cli"` in `create_provider()`
- `Settings.toml`: Example config switch (provider = "claude-cli")

### Usage

```toml
[ai]
provider = "claude-cli"
model = "claude-opus-4-6"
```

Requires `claude` CLI installed and authenticated. No API key needed.

## Test plan
- [x] `cargo check` passes
- [ ] Run a review with `provider = "claude-cli"` and verify findings are produced
- [ ] Verify timeout triggers correctly on a hung process
- [ ] Verify graceful handling when `claude` is not installed

### DIFF
--- a/src/ai/claude_cli.rs
+++ b/src/ai/claude_cli.rs
@@ -125,7 +125,7 @@ impl AiProvider for ClaudeCliProvider {
 
 /// Build the full text prompt from the AiRequest.
 /// Embeds system prompt, conversation history, tool definitions, and instructions.
-fn build_prompt(request: &AiRequest) -> String {
+pub fn build_prompt(request: &AiRequest) -> String {
     let mut out = String::new();
 
     // System prompt
@@ -214,7 +214,7 @@ fn parse_usage(outer: &Value) -> Option<AiUsage> {
     })
 }
 
-fn parse_inner_response(text: &str, usage: Option<AiUsage>) -> Result<AiResponse> {
+pub fn parse_inner_response(text: &str, usage: Option<AiUsage>) -> Result<AiResponse> {
     // Try extracting JSON (might be in a markdown code block)
     let json_str = extract_json(text);
 
--- a/src/ai/codex_cli.rs
+++ b/src/ai/codex_cli.rs
@@ -1,0 +1,146 @@
+// Copyright 2026 The Sashiko Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


paf0186:
Just a note on justification:
Within the subscription limits, Claude Code Max is something like 10x cheaper than the API.

rgushchin:
If you can add Gemini CLI and OpenAI codex, it's even better. Sashiko's approach is to support all models which can produce actually useful results.

paf0186:
So I'll put a little human on the front of this, but my Claude and I have finished up and tested the claude code CLI provider and tested it out.  Required a few other fixes.  I'm going to have it take a stab at Gemini and Codex a bit later.

"Updated — three commits now:

0567650 feat: add claude-cli provider — core provider (src/ai/claude_cli.rs), shells out to claude --print
451456e settings: restore upstream defaults — gemini stays default, claude-cli documented as commented-out alternative
1140116 fix: pass correct stdio provider to review child — the review binary was always started with --ai-provider stdio-gemini regardless of config. Now maps claude/stdio-claude/claude-cli → stdio-claude. This affects which stdio protocol the child uses, and the reported context window size (200K vs 1M).
Benchmark result (1 review from benchmark_tiny.json, btrfs bitmap refactor):


status=Reviewed  model=claude-haiku-4-5-20251001  provider=claude-cli
tokens_in=182  tokens_out=53,029  tokens_cached=436,943
findings: 0
Full pipeline works end-to-end: claude-cli → stdio proxy → review child (9 stages) → reviewed. Zero findings is expected for Haiku on a low-risk refactor. Running the full tiny benchmark next.

Note: all providers currently hardcode context_window_size rather than deriving it from the model. This is a pre-existing issue (not introduced here) — worth addressing separately since max_input_tokens in settings is a budget cap, not the model's actual context window. A model lookup table or new setting would be the right fix.

Fix 3 affects all providers, not just claude-cli — happy to split it into a separate PR if preferred.

Re: Gemini CLI and Codex CLI providers — I have working invocation patterns and can add those in a follow-on."

paf0186:
Note w/the 0 findings - I used the smallest model, Haiku, to save usage but confirmed it did a real review.  It's just not a very good reviewer is all.

PR #53, opened 2026-03-26
cli: simplify server URL handling

ackerleytng:
Add a default value for the server argument in the Cli struct and remove the logic to load settings for the URL. This removes an unnecessary dependency on Settings in the CLI and allows sashiko-cli to be run from any directory, even if it doesn't contain a Settings.toml file.


derekbarbosa:
Hi,

Sorry for letting this go under the radar. There has been considerable divergence since your PR opened, would you like to rebase + update it?

PR #55, opened 2026-03-26
RFC: Local submission

ackerleytng:
The problem I was facing was that I was passing a local directory (the directory is not a git remote as in https://stackoverflow.com/q/6648995) for `--repo`, and the server was instead reviewing `HEAD~5..HEAD` from the "local" repository configured in settings as `repository_path`. This was confusing and didn't quite match what the docs was suggesting.

This change proposes to formalize `--repo` in the cli (probably should be renamed `--remote`) as the remote that contains commit/range to align with what I understand to be the current server behavior. The current server behavior will interpret `HEAD~5..HEAD` according to sashiko's repository, which is not the same as what was passed from `--repo`.

Hence, these changes resolve the references client-side, if `--repo` is passed, before sending it server side. Server side validation is missing for this RFC.

I think a better alternative would be to formalize the different references in the API and redefine `--repo` for the cli:

1. repo (regular directory, not git remote) - sashiko does no fetching, repo must be up to date
2. git remote - sashiko will fetch from the provided remote and interpret references based on that
3. sashiko's repository - no fetches, sashiko assumes its own repo is up to date

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ cargo run --bin sashiko-cli -- [COMMAND]
   - If `INPUT` is omitted and stdin is a terminal, it defaults to the current `HEAD`.
   - `--repo` (or `-r`) specifies an external git repository (URL or local path acting as a git remote).
   - If `--repo` is provided, the CLI resolves references to full 40-character SHAs locally and verifies their existence in the remote before submission.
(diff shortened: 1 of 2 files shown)


derekbarbosa:
@rgushchin this has been resolved since the introduction of the new local review mechanism, right?

PR #75, opened 2026-04-03
RFC: api.rs: init some rustdocs

derekbarbosa:
This PR is hopefully the first in a series to introduce & implement module-level documentation.

This PR aims to introduce comprehensive rustdocs for src/api.rs

Specifically, it targets documentation for the central API surface, documenting the intended purpose of the code, and providing useful insight into various types/structs/methods/impls.

Sample output of old `cargo docs --no-deps --open`:
Screenshot From 2026-04-03 12-19-04
Screenshot From 2026-04-03 12-19-22

New sample output:
Screenshot From 2026-04-03 12-26-03
Screenshot From 2026-04-03 12-33-41

Prompt provided to Gemini-CLI with latest gemini-pro model in Plan Mode to produce code comments:

```
Gemini,

The goal of this session is to improve the module-level documentation of this project. As it stands, there aren't any rustdoc comments, inline examples, or annotations of functions.

We're going to start documenting this code module-by-module.

First, ingest the GEMINI.md file at the root of the project repository.

Then, run a @generalist agent to fetch and retrieve what are considered best practices for rust code documentation.

Additionaly, run a @codebase_investigator agent in parallel to read the api.rs module, and examine what the current gaps in documentation are.

You are to examine the code with a high level of scrutiny. Code comments and docs should _never_ describe *WHAT* the code does, but rather the underlying **WHY**

Most of this code is very self-documenting as-is, we are simply enabling ease of navigation via rustdoc, and describing the intended purpose of traits, methods, structs, etc. Comment overuse will be outright rejected.
```

### DIFF
--- a/src/ai/claude_cli.rs
+++ b/src/ai/claude_cli.rs
@@ -33,7 +33,9 @@ use tokio::process::Command;
 use tokio::time::timeout;
 use tracing::{debug, warn};
 
-use crate::ai::{AiProvider, AiRequest, AiResponse, AiRole, AiUsage, ProviderCapabilities, ToolCall};
+use crate::ai::{
+    AiProvider, AiRequest, AiResponse, AiRole, AiUsage, ProviderCapabilities, ToolCall,
+};
 
 pub struct ClaudeCliProvider {
     pub model: String,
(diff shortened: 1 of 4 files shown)


rgushchin:
Sorry, I'm not sold yet on adding auto-generated documentation for the llm-generated code. Maybe a bit later once the whole project will be more mature.

PR #79, opened 2026-04-06
.github/workflows: Init CI

derekbarbosa:
.gitlab/workflows: Init CI

Initialize some simple CI for the repository. Specifically, set up the
following workflows:

1. Setup a DCO check
2. Set up basic linting (yamllint, cargo-{clippy,fmt})
3. Run cargo test

The workflows follow a modular approach for easier maintenance and
modification + reuse.

Also, fix basic formatting errors to prevent errors on future PRs.

Signed-off-by: derekbarbosa <derekasobrab@gmail.com>

PR #82, opened 2026-04-09
.github/workflows: remove the --quiet arg in cargo-fmt

derekbarbosa:
Having CI fail due to a formatting error can be problematic when said
error is not propagated to the user.

Break up the rust potion of the linting phase into two sections, and
remove the quiet argument so that it shows up in the workflow logs.


derekbarbosa:
see: https://github.com/sashiko-dev/sashiko/pull/79

PR #86, opened 2026-04-12
db: Fix database insertion race condition

jpoimboe:
When multiple operations run concurrently, last_insert_rowid() can return the ID from a different operation's insert, causing data to be associated with the wrong record.

For reviews, this results in the following:

  1. The affected review's output is overwritten with results for a different commit

  2. The review that should have received the output is left permanently stuck at "Pending" with no output

The problem is that INSERT and SELECT last_insert_rowid() are done as separate operations, allowing another concurrent insert to interleave between them and change the returned ID.

Fix it by combining the insert and ID retrieval into a single INSERT ... RETURNING id statement.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -566,10 +566,11 @@ impl Database {
         baseline_id: Option<i64>,
         prompts_hash: Option<&str>,
     ) -> Result<i64> {
-        self.conn
-            .execute(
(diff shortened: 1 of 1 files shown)


rgushchin:
LGTM. Thank you!

PR #95, opened 2026-04-15
api: fix API endpoint for fetching patchset data

### LINKED ISSUE #57: Move `reviews.[].logs` from `api/patch` to a dedicated endpoint
Hello,

I noticed that it was quite slow to load some pages when commuting, with a slow connection.

It would be nice to move the review logs from the `/api/patch` endpoint to a dedicated one: maybe there is no need to load a few MBytes that are not rendered on the UI (except via the dedicated `/log/` route)? :)

matttbe:
The /api/patchset endpoint has been recently introduced to provide a lighter view, without the full review log and input context for every patch, which can be huge and not used in the patchset detail view.

Except that it has been only changed for the baseline log view, but not the patchset detail one. The content is the same as in /api/patch, only with trimmed log info, so simply switching to the new endpoint fixes this.

Fixes: 3dbb448ed0a4 ("api: introduce /api/patchset and /api/review_log endpoints")
Closes: #57

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -1604,7 +1604,7 @@
         async function renderPatchsetView(id, query = {}) {
             const page = parseInt(query.page) || 1;
             const limit = parseInt(query.limit) || 50;
-            const res = await fetch(`/api/patch?id=${encodeURIComponent(id)}&page=${page}&limit=${limit}`);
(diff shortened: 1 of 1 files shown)


rgushchin:
Thank you, good catch! Can you please add your Signed-off-by line?

matttbe:
Sorry, I forgot that, just added with the right SoB and author

rgushchin:
Merged, thanks!

PR #97, opened 2026-04-16
Update README with "just" targets and needed prerequisites

igoropaniuk:
The Justfile has been the project's task runner but was not documented in the README. New contributors
had no way to discover available dev commands without reading the Justfile directly.

Also add system package prerequisites.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,25 @@ git commit -s
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


derekbarbosa:
`just` has been migrated to `make` for compatibility

PR #104, opened 2026-04-20
k8: add linux-renesas-soc and forward reviews to me

wsakernel:
I will evaluate a workflow with me sending out curated Sashiko reviews, so send the Sashiko reports to me only first.


ukleinek (commented):
$Subject ~= s/k8/k8s/ maybe?

PR #105, opened 2026-04-20
k8s: add linux-omap mailing list

khilman:


rgushchin:
Merged manually. Thank you!

ukleinek:
I wonder if it would make sense to make the list of tracked mailing lists a (sorted) yaml list. That probably reduces this kind of conflict.

Issue #107, opened 2026-04-20
Claude: Remote AI Error: Invalid request: input should be 'object'


khilman:
Attempting a local setup with Claude CLI as describe in the docs.  I'm using an `ANTHROPIC_API_KEY` that works with other tools, and getting the following errors.  

Using Claude Code v2.1.114, which also works fine with this API key.

```
2026-04-20T21:44:19.110077Z  INFO sashiko::ai::claude: Claude response received. Tokens: in=5395, out=318, cached=0
2026-04-20T21:44:19.110785Z  INFO sashiko::reviewer: [review-bin] Phase 0 JSON parse error: expected value at line 1 column 1
2026-04-20T21:44:19.115826Z  INFO sashiko::reviewer: [review-bin] Running planning pre-phase
2026-04-20T21:44:22.253594Z  INFO sashiko::ai::claude: Claude response received. Tokens: in=133153, out=6, cached=0
2026-04-20T21:44:22.254139Z  INFO sashiko::reviewer: [review-bin] Planning phase JSON did not contain 'relevant_stages' array
2026-04-20T21:44:22.254178Z  INFO sashiko::reviewer: [review-bin] Running Stage 1
2026-04-20T21:44:22.684235Z  INFO sashiko::reviewer: [review-bin] Stage 1 AI execution failed (inner attempt 1/3): Remote AI Error: Invalid request: {"type":"error","error":{"type":"invalid_request_error","message":"tools.0.custom.input_schema.type: Input should be 'object'"},"request_id":"req_011CaFkh1mprCTxTQkZXJsqY"}
2026-04-20T21:44:22.878604Z  INFO sashiko::reviewer: [review-bin] Stage 1 AI execution failed (inner attempt 2/3): Remote AI Error: Invalid request: {"type":"error","error":{"type":"invalid_request_error","message":"tools.0.custom.input_schema.type: Input should be 'object'"},"request_id":"req_011CaFkh3JrVzViHGqtweJfQ"}
2026-04-20T21:44:23.080141Z  INFO sashiko::reviewer: [review-bin] Stage 1 AI execution failed (inner attempt 3/3): Remote AI Error: Invalid request: {"type":"error","error":{"type":"invalid_request_error","message":"tools.0.custom.input_schema.type: Input should be 'object'"},"request_id":"req_011CaFkh49DBnEz3vE4TnBBj"}
[...]

PR #108, opened 2026-04-20
WIP: DONOTMERGE: example fixes for Claude API

khilman:
Was getting errors when trying to use Claude with $ANTHROPIC_API_KEY. Debugging with Claude Code, it found several issues:

1. Tool schemas (src/worker/tools.rs) — All type names were uppercase ("OBJECT", "STRING", "ARRAY", etc.), which Gemini accepts but the Claude API rejects with: "tools.0.custom.input_schema.type: Input should be 'object'" Changed all to lowercase. This was blocking every Stage 1 attempt.

2. Phase 0 and Planning schemas (src/worker/prompts.rs) — Same uppercase type issue in the inline JSON schemas used for those phases.

3. Claude ignores response_format (src/ai/claude.rs) — When AiResponseFormat::Json is requested, the Claude client was sending nothing special, so Claude responded with prose instead of JSON. Fixed by appending an explicit instruction to the system prompt when JSON format is requested, including the full schema so Claude knows the exact key names required.

   Two sub-issues fixed here: a. When system prompt is None (as in the planning phase request), the original fix never fired. Now a minimal system block is created to carry the JSON instruction in that case. b. The planning prompt example showed a bare array "[4, 5, 6, 7]" but the parsing code expected {"relevant_stages": [...]}. Updated the example to match the key name the code looks for.

After these fixes, all three phases work correctly with Claude:
  - Phase 0 selects relevant subsystem guides
  - Planning phase selects relevant review stages
  - Stage 1+ tool calls succeed

The changes here are entirely generated by Claude Code, and may break Gemini since the type casing conventions differ.  This patch is only intended as an example of what the Claude API expects.

### DIFF
--- a/src/ai/claude.rs
+++ b/src/ai/claude.rs
@@ -248,11 +248,38 @@ fn translate_ai_request(request: &AiRequest, enable_caching: bool) -> Result<Cla
     let mut system_blocks = Vec::new();
 
     // Extract system prompt from the explicit system field
+    let json_instruction_owned: Option<String> =
+        if let Some(crate::ai::AiResponseFormat::Json { schema }) = &request.response_format {
+            // Claude has no native JSON mode; enforce it via instruction.
+            if let Some(schema) = schema {
+                Some(format!(
+                    "\n\nYou MUST respond with valid JSON only matching this schema: {}. Do not include any explanation, markdown, or code fences — output raw JSON.",
+                    serde_json::to_string(schema).unwrap_or_default()
+                ))
+            } else {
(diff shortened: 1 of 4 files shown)


khilman:
Submitted by request from @rgushchin in https://github.com/sashiko-dev/sashiko/issues/107

This is not ready for merge as it seems it will likely break Gemini support.  Submitted as an example of how to get Claude working (and entirely written by Claude.)

khilman:
Added a couple more fixes that make Claude CLI work so it can be used with Pro/Max/Team plans that don't use API key.

khilman:
Seems the Claude API changes here are no longer needed.  Someone else has now fixed these in master.  However, the claude-cli changes (2nd patch) seem still relevant.  I'll rework as a separate PR.

PR #110, opened 2026-04-21
settings: support Settings.local.toml for local overrides

kuba-moo:
Either I'm confused of something strange happened to the config. It's committed and in the .gitignore.
I think it makes sense to have the config / sample config committed so let's add a secondary "local" config for overrides?
Sorry if I'm completely off base..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,5 @@ third_party/linux/
 benchmarks/smoke-repo/
 review_trees/
 sashiko.log
-Settings.toml
+Settings.local.toml
 email_policy.toml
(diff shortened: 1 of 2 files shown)


rgushchin:
It's correct: I often have to change settings locally and gemini tends to commit these temp changes with random commits, adding it to .gitignore prevents this. In a rare case when you actually want to update it, it can be done manually.
I'm not sure I understand the point of the second config, can you, please, elaborate? We can totally optimize the default config to the local usage.

kuba-moo:
Let me rephrase to make sure i understand -- the .gitignore entry prevents Gemini from committing the changes? Doesn't seem to work as well for Claude :S
The local config is _my_ workaround for Claude committing the config changes as part of random commits :) Also I was hacking on the same tree simultaneously on two machines, one trying to improve the claude CLI the other on Bedrock, and I needed two different configs.

PR #111, opened 2026-04-21
prefetch: use tree-sitter AST for type and definition resolution

kuba-moo:
My first attempts to run Claude showed that there's a lot of exploration of the types. Looking closer that the compiled context revealed that there was all sort of false positives hits in there instead of the relevant information. Things which didn't matter, stuff from tools/ that matched the regexps. IDK if this is something with my setup or reproducible, but this commit fixes it. The context looks sensible. I didn't notice any slowdown in the preparation stage. The number of msgs sent to the LLM dropped by ~20% since the context is now correct. 

Fully vibe-coded, LMK if I should understand what this does..

### DIFF
--- a/src/worker/prefetch.rs
+++ b/src/worker/prefetch.rs
@@ -1,9 +1,9 @@
 use anyhow::Result;
 use regex::Regex;
 use std::collections::{HashMap, HashSet};
-use std::path::Path;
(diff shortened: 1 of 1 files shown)


rgushchin:
This looks great, thanks! But let me run the full test, I'm interested if it bring any improvement.

PR #112, opened 2026-04-21
email: add rust-for-linux subsystem to policy

ojeda:
In Rust for Linux, we would like to at least get the reports sent to the original author, thus configure sashiko.dev to do so. In addition, Cc me as well.

Depending on how it goes, I may request later to just send the reviews to the mailing list.

Thanks!

### DIFF
--- a/sashiko.dev/email_policy.toml
+++ b/sashiko.dev/email_policy.toml
@@ -56,3 +56,8 @@ cc = ["tj@kernel.org", "hannes@cmpxchg.org", "mkoutny@suse.com"]
 [subsystems.renesas-soc]
 lists = ["linux-renesas-soc@vger.kernel.org"]
 cc = ["wsa+renesas@sang-engineering.com"]
+
+[subsystems.rust-for-linux]
+lists = ["rust-for-linux@vger.kernel.org"]
+reply_to_author = true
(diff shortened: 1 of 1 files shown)


Lyude:
Hey - with all due respect, could we maybe make this an opt-in instead of adding it for the whole mailing list and maybe not expand this to sending out reviews to the entire mailing list?
I really don't find this tool useful, and was not terribly happy to find it suddenly responding to my patches this morning. It sends an incredible volume of comments and so far every time I've looked at what it's sent me, the huge majority of these comments are wrong and I end up losing time having to go and verify all of them. If they suddenly are getting sent to the entire rust for linux mailing list, it concerns me that suddenly the onus of going through the huge number of false positives this tool generates will still end up on my lap because people may misinterpret them to be actual correct review comments that need addressing - even if I've blocked the tool myself. I don't really think that's fair, and I'd much rather people who do find these tools useful be the ones to take responsibility for using them and verifying their output.
Thank you!

rgushchin:
Hi! I consider Sashiko the tool for maintainers and fully rely on corresponding maintainers decision on how they are willing to use it or not use at all. Please, discuss it with rust-for-linux maintainers, I’ll respect the consensus decision.

PR #115, opened 2026-04-22
claude: add more settings to claude client

mswiatko:
Add max_tokens, base_url, thinking and effort. Follow the scheme from other clients (Bedrock, OpenAI, etc).

As thinking can be set now, there is a need to override the temperature when the thinking isn't disabled. In this case claude is accepting only 1.0 as a temperature.

Set default max_tokens and url to the previous hardcoded values (4096 and antrophic url).

Default thinking is None, however, most models (probably) treat request without thinking as default thinking (enable or adaptive). Because of that temperature is set to request value only when the thinking is set to disable by settings.

### DIFF
--- a/src/ai/claude.rs
+++ b/src/ai/claude.rs
@@ -87,6 +87,10 @@ pub struct ClaudeRequest {
     pub tools: Option<Vec<ClaudeTool>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub temperature: Option<f32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub thinking: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub effort: Option<String>,
 }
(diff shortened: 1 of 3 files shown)


kuba-moo:
Makes sense FWIW. I only added this to bedrock, but we need the same thing in antrhopic.
One difference is that AFAICT you're sending temp of 1 always with thinking, and I think I was just skipping it entirely.

mswiatko:
> Makes sense FWIW. I only added this to bedrock, but we need the same thing in antrhopic. One difference is that AFAICT you're sending temp of 1 always with thinking, and I think I was just skipping it entirely.

Right, from the documentation:
"Deprecated. Models released after Claude Opus 4.6 do not support setting temperature. A value of 1.0 of will be accepted for backwards compatibility, all other values will be rejected with a 400 error."

I will add this change in my other PR, thanks.

PR #116, opened 2026-04-22
build: make bedrock provider an optional feature

kuba-moo:
Building sashiko OOMs my laptop :( Roman suggests that the bedrock dependency is a major source of slowness.

Gate bedrock and it's dependencies, but keep it enabled by default to minimize changes.
I'm slightly tempted to disable it by default, but IDK. Could be just me that is annoyed by it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,12 +4,16 @@ version = "0.1.4"
 edition = "2024"
 default-run = "sashiko"
 
+[features]
+default = ["bedrock"]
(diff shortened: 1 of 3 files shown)


rgushchin:
Did it help with OOMing? If yes, I'm all in on this, it was slowing down the build significantly.

kuba-moo:
Not an MM expert, but previously I saw OOM killer invoked. During last two builds (sample of 2 - very scientific) with bedrock disabled I saw occasional slowdowns but no freezes and full build completes in 4m 28s.

PR #118, opened 2026-04-22
prompts: add pipeline scoping to prevent full-investigation per stage

kuba-moo:
Claude has a tendency to do the full investigation at every stage. The tweaks and repeats in this PR seem to stop it _most_ of the time. Stage 2 (Implementation Verification) is still prone to going off and doing full analysis about half of the time, but at least the other stages are fairly reliably constrained.

### DIFF
--- a/src/worker/prompts.rs
+++ b/src/worker/prompts.rs
@@ -145,15 +145,12 @@ impl PromptRegistry {
             current_date
         );
 
-        content.push_str(&date_fact);
-        content.push_str("You are an expert Linux kernel maintainer. Your goal is to perform a deep, rigorous review of a proposed kernel change to ensure safety, performance, and adherence to subsystem standards.\n\n");
-        content.push_str("TOOL USAGE: When you need to gather information using tools, actively batch parallel or independent tool calls into a single response to minimize the number of conversation turns.\n\n");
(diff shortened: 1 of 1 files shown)


kuba-moo:
Ah, should have mentioned, this also coincidentally makes local Sashiko running on claude code usable for me. Otherwise each stage times out trying to go too deep. The results, however, are rather mixed :s Claude code + look at Chris's prompts outperforms in terms of issues found.

rgushchin:
Yeah, I’m afraid it might cause a significant discovery rate drop. In many cases you just need to burn more tokens to get better results. I bet if we run each stage 3 times and then have a separate stage for verification of each bug, results will be superior.
I can’t merge it without a full scale experiment (1k bugs) and I’m a bit out of quota to run it in next few days :(
If you can’t run it yourself (with and without your change), then I’ll postpone it a bit and bundle with some other changes for an over-a-weekend test.

kuba-moo:
Understandable, no worries. In my testing Claude is not usable without this change. Meaning - I can't really run the full regression :( Claude burns $30-$40 per patch before this, after $5-8 depends on how (un)lucky we get with the double-detection.

derekbarbosa:
hey @kuba-moo -- it isn't ideal, but maybe editing the existing prompts and supplying them as configuration as described in #141 may help in your use cases?

kuba-moo:
The insert in this PR is minimal. I think #141 diverges too much for my needs. IOW I prefer to stay in sync with the upstream with an occasional rebase config than completely branching off to custom prompts.

PR #131, opened 2026-04-28
Api submit mbox batch

helq:
We made these changes to fix an issue where patches submitted in a single batch via the API were split into multiple incomplete patchsets due to large timestamp differences or distinct authors.

To solve this, we grouped patches by a common synthetic submission ID. To implement this robustly and avoid brittle content-sniffing heuristics (like checking for '@' to distinguish synthetic IDs from real Message-IDs), we introduced an explicit  MessageSource  enum to track the origin of each message. This ensures correct ID resolution and appropriate author strictness based on the source.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use crate::db::Database;
-use crate::events::Event;
+use crate::events::{Event, MessageSource};
 use crate::fetcher::FetchRequest;
 use crate::settings::ServerSettings;
 use axum::{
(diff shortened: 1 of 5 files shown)


src/api.rs
```diff
@@ -334,6 +334,7 @@ async fn submit_patch(
 
             let event = Event::RawMboxSubmitted {
                 raw,
+                article_id: id.clone(),
```
rgushchin: can it be patchset_id? Or something like this... Or maybe submission_id, something like this. I'm not sure what article means hre.

helq: I went for "article_id" because it is the ID name that the rest of the code uses at ingestion time. I didn't want to use "patchset_id" for this because "patchset_id" is managed by the database which we hit only once we have assembled the batch. I think you are correct, "submission_id" is much cleaner though. I am pushing a new commit where I renamed article_id to submission_id for mbox submissions. That makes it clear that the id refers to the submission, not to an email

derekbarbosa (commented):
I started a review and accidentally left it in pending for some time. Apologies!

@rgushchin this would resolve some of the issues we're seeing with developers queuing up their patch-sets to redhat's "remote" instance prior to sending to upstream mailing list :-)

helq:
This work has been superseeded by #411

PR #132, opened 2026-04-28
db: improve patchset matching by enforcing version and index checks, and allowing cross-thread merges for complementary series 

captain5050:
This pull request introduces two critical improvements to the patchset matching and merging engine in `Database::create_patchset` to guarantee accurate series isolation across versions and threads:          
  
#### 1. Strict Candidate Validation in `cover_letter_message_id` Matching Loop

Fixes a bug where distinct versions (e.g., v1 vs v2) or colliding patch indices were incorrectly merged into the same patchset when matched by `cover_letter_message_id`. 
 - Enforces **version compatibility checks** to prevent a v2 series from being merged into a v1 series. 
 - Enforces **index collision prevention** to ensure that if a candidate already contains a patch for the same part index, it is skipped.
  
#### 2. Relax Thread Isolation for Incomplete Complementary Patchsets

Addresses the problem where resending the remaining patches of a large series (e.g., due to a `git send-email` failure) in a new thread created a separate incomplete series.
 - Relaxes the `thread_compatible` check to allow cross-thread matching when the candidate series is incomplete (`existing_received < existing_total`) and there is no index collision.
 - Prioritizes same-thread matching candidates over cross-thread matches during sorting.
 - Prevents cross-thread merging of distinct existing patchsets in the multiple matches merge loop to avoid accidental patch stealing.
  
Both changes have been fully verified with the new `tests/cover_letter_version_merge_test.rs` and `tests/cross_thread_complementary_merge_test.rs` integration tests. All existing tests are fully passing.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1862,8 +1862,10 @@ impl Database {
             };
 
             // Thread Enforcement: To prevent cross-thread "stealing" of patches for resends of the same series,
-            // we strictly require multi-part series patches to belong to the same thread.
-            let thread_compatible = same_thread || is_singleton;
+            // we strictly require multi-part series patches to belong to the same thread, UNLESS
+            // the candidate series is incomplete, in which case we allow merging complementary parts.
+            let thread_compatible =
+                same_thread || is_singleton || (existing_received < existing_total);
 
             if author_or_series_match
                 && (!strict_author || (date - existing_date).abs() < 86400)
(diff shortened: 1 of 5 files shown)


rgushchin:
Hey! Thanks for the PR! How it was tested? I suggest running it with --no-ai and pulling like 100k or 1M messages from various mailing lists and see how it will perform. I've seen many examples when the version tag is set only on the cover letter or not set on the cover letter.

PR #135, opened 2026-04-28
Feat remote baseline

helq:
Currently,  sashiko  only relies on upstream trees listed in the  MAINTAINERS  file to find baseline candidates for applying patches. This causes patches targeting internal or downstream repositories to fail to apply. We need a way to specify custom repositories and check all or specific branches as potential baselines.
  
  • Configurable custom remotes: Added support in  Settings.toml  to define a list of custom remotes with options to check all branches or a specific subset.
  • Dynamic baseline injection: Updated the baseline resolution logic to fetch these custom remotes and inject their branches as candidates when trying to apply a patch.   
  • Optimizations and hardening: Added deduplication of candidates by commit SHA to prevent redundant application tests, and isolated unit tests using temporary git repositories to avoid polluting the developer environment.

### DIFF
--- a/Settings.toml
+++ b/Settings.toml
@@ -46,6 +46,13 @@ port = 8080
 [git]
 repository_path = "third_party/linux"
 
+# Track custom remotes and attempt to apply patches on their branches
+# [[git.custom_remotes]]
+# name = "remote-name"
+# url = "git://my-custom.com/remote"
+# check_all_branches = true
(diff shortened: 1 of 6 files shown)


helq:
Changes in my branch were not reflected here. I opened a new PR https://github.com/sashiko-dev/sashiko/pull/138

PR #139, opened 2026-04-29
git_ops: Fix worktree creation race condition

jpoimboe:
Concurrent git worktree add calls race on the shared .git/worktrees/ directory.  Git worktree creation is not atomic -- it creates the directory and writes HEAD, commondir, gitdir files sequentially.  A concurrent git worktree add can enumerate existing entries and read a partially-initialized one, causing:

  fatal: failed to read .git/worktrees/.../commondir: Success

Add a global async mutex around git worktree add, remove, and prune operations, following the existing get_global_config_lock() pattern.

### DIFF
--- a/src/git_ops.rs
+++ b/src/git_ops.rs
@@ -68,20 +68,42 @@ impl GitWorktree {
 
         info!("Creating worktree at {:?}", path);
 
+        // Split worktree creation into two phases:
+        // 1) metadata update (under lock, fast)
+        // 2) file checkout (no lock, parallelizable)
+        let output = {
+            let lock = get_worktree_lock();
+            let _guard = lock.lock().await;
+            Command::new("git")
+                .current_dir(repo_path)
+                .args(["-c", "safe.bareRepository=all"])
+                .arg("worktree")
+                .arg("add")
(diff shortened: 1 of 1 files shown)


rgushchin:
Hi Josh! Thanks for the PR! I think the problem is real, but the fix might be problematic: sashiko.dev is reviewing 50 patch(set)s in parallel and synchronizing all git operations might bottleneck it on i/o. What I believe can be done instead: we can split the operation in two parts: metadata modifying (and synchronize it with a mutex, as you suggested) and file checkout, which should be ok to keep asynchronous.

jpoimboe:
@rgushchin I could be missing something, but the intention of this patch is only to use the lock for creating/removing/pruning a worktree, not for all git operations.  Presumably that only happens when starting or ending a review?

rgushchin:
Yes, but if I read it correctly checking out the new worktree will be under the lock, right? If you need to checkout 50 trees it's significantly slower

jpoimboe:
Ah, gotcha.  Yeah, that makes sense.

jpoimboe:
v2: move out the slow parts of the worktree operations outside the lock

PR #147, opened 2026-05-04
add batman(-adv) mailing list to recognized mailing lists

ecsv:
I would like to request to get the batman-adv mailing list added to the email policy, CCing only maintainers. This is necessary to not get in conflict with the net 24h embargo from commit 2ece30494cda ("Implement AI review embargoes for patchsets"). This embargo was added to avoid patchset authors to send more than 1 version per 24 hours.

All maintainers were contacted and agreed to get their addresses in the Cc for reviews of patches to b.a.t.m.a.n@lists.open-mesh.org

* @lindnermarek
* @simonwunderlich
* @ordex

---

The batman(-adv) lore id was added because the batman-adv patches are only sometimes reviewed via the netdev mailing list.  It can easily happen that they are send for a first review only to the batman(-adv) mailing list. It would be nice to also get the reviews for these patches directly. Otherwise, we would maybe only see these reviews after the patches are applied to the batman-adv tree and they got submitted to netdev as PR.


rgushchin:
It won't help with embargoes as of now, but I'm going to implement a feature to opt out soon.
Thanks

PR #151, opened 2026-05-05
email: Add email policy for Devicetree

robherring:


rgushchin:
I am a bit concerned about the blast radius: it looks like many patches targeting e.g. net are cc'ing the devicetree mailing list. Maybe let's start with something a bit more conservative than reply_all? E.g. reply to the devicetree mailing list, author and a couple of selected people?

robherring:
The problem with just the DT maintainers and list is most patches go thru subsystem trees and subsys maintainer may not see the sashiko report nor answers to it. Most times maintainers wait for our ack, but not always.

But sure, we can start with just DT maintainers and list and see how it works.

PR #157, opened 2026-05-07
Fetcher improvements

helq:
These patches introduce multiple improvements for the fetching algorithms:

- Fetching restarts when we restart sashiko.
- If we receive multiple api/submit requests for the same SHA, we only create one entry.
- Retry logic for fetching, as the remotes might not be completely stable.

These changes are made on top of https://github.com/sashiko-dev/sashiko/pull/131


helq:
Superseeded by the durable fetcher queue #418

PR #158, opened 2026-05-07
docker: prune stale git worktrees synchronously at startup

helq:
Prune stale git worktrees in the entrypoint script before Sashiko starts. This prevents broken worktree refs from causing git fetch failures when GKE pods are evicted or restarted.


rgushchin:
Please not add more functionality to docker: let's make sashiko work on it's own. I want as little docker as possible, ideally 0 in the long run.

PR #160, opened 2026-05-08
fix: support both abab7 and MiniMax model prefixes

Dwyane-Yan:
Hi Dear Maintainers,

Thank you for your great work on this project! I'm submitting this PR to improve compatibility with MiniMax's recent model releases.

**Context:**
MiniMax has recently introduced new models (like MiniMax-M2.7 [1] and MiniMax-Text-01 [2]) using the `MiniMax-` prefix, alongside the legacy `abab7-` prefix models.

**The Issue:**
Previously, only the "abab7-" prefix was checked to route requests to the MiniMax API endpoint. This caused models with the new "MiniMax-" prefix to fall back to the default OpenAI endpoint, resulting in request failures.

**The Fix:**
This commit updates the prefix matching logic to recognize both "abab7-" and "MiniMax-" prefixes, ensuring correct API routing and context window assignment for both legacy and new MiniMax models. I have also updated the related design documentation (`DESIGN_OPENAI_COMPAT_PROVIDER.md`) to reflect this change.

Thanks for your time and consideration!

[1] https://minimaxi.com/news/minimax-m27-zh
[2] https://platform.minimaxi.com/document/ChatCompletion?key=667bdde33be2027f69b71d4c

### DIFF
--- a/designs/DESIGN_OPENAI_COMPAT_PROVIDER.md
+++ b/designs/DESIGN_OPENAI_COMPAT_PROVIDER.md
@@ -16,7 +16,7 @@ The official OpenAI API uses `max_completion_tokens` in the request body (introd
 | Stdio support | Not needed for OpenAI-compatible provider |
 | Thinking/reasoning support | Not included in initial implementation (`thought: None` always) |
 | Temperature | Always passed through from `AiRequest` when present |
-| URL configuration | `base_url` from settings → model-based default (glm-*, moonshot-*, abab7-*, others) |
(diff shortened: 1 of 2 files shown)


Dwyane-Yan:
As a Linux kernel contributor, I really appreciate this project. I specifically want to express my gratitude for Sashiko—it is a fantastic tool that actually helped me catch an issue in one of my recent kernel patches! 

Recently, I have been trying to deploy Sashiko locally, and I ran into a routing issue during the setup process, which led to this PR. 

Since this is my first contribution to the Sashiko community, please kindly overlook any missteps on my part. I am very open to learning and would greatly appreciate any guidance or corrections if there's anything I should do differently.

PR #164, opened 2026-05-09
Claude show needs fixes

acmel:
One cset addresses the crash concerns from Derek on the original claude show needs PR, the other makes the --diff recently added feature to respect the --color settings.

This branch was force pushed after a rebase.


derekbarbosa:
LGTM, ty!

PR #166, opened 2026-05-10
AI: classify AI provider retry errors to handle Rate Limit responses

jgunthorpe:
We have an OpenAI type LLM proxy that generates rate limit responses. Currently Sashiko just ignores the required delay and immediately resends the request, which gives another rate limit and eventually it blows up the whole review.

After this patch the rate limiting works more like I would expect:

 2026-05-10T13:30:55.155717Z  WARN sashiko::ai::openai: OpenAI 429 Rate Limit. Retry in 60s
 2026-05-10T13:30:55.155946Z  WARN sashiko::ai::quota: Quota exhausted! Blocking all LLM requests for 60.00s
 2026-05-10T13:30:55.155967Z  INFO sashiko::ai::quota: [ps:21 p:1 s:3] Global AI rate limit/quota active. Waiting for 60.00s...
 2026-05-10T13:31:55.158072Z  INFO sashiko::ai::openai: Sending OpenAI request...

I have to say I'm a little unconformable sending this, I don't know rust and the entire thing was written by Claude and Codex as a reviewer/implementer pair. I've tested it and confirmed it improves my situation. The AI generated plan that created the commit is in the commit message. I didn't think this was significant enough to qualify for a designs/ file?  I challenged the AI a bit and it seems to think this is a Rusty pattern. Though Claude did offer an alternative:

```
2. Unified enum with #[from]

#[derive(thiserror::Error)]
pub enum AiError {
    #[from] Claude(ClaudeError),
    #[from] OpenAi(OpenAiCompatError),
    #[from] Gemini(GeminiError),
    Wrapped(String),
}
Every provider call returns Result<_, AiError>. The classifier becomes a single exhaustive match — no downcasting at all.

This is the most "Rust-y" answer and is closest to what your "class hierarchy" instinct is reaching for. But the cost is high here: you'd have to change return types throughout src/ai/, src/reviewer.rs, the proxy, the benchmark, and the worker. Every ? that crosses the boundary needs the conversion. Given that the whole codebase is on anyhow, you'd be swimming upstream.
```

But it sounded scary so I left it. As far as I can tell the AI tool ran all the requested checks and they worked.

### DIFF
--- a/designs/DESIGN_AI_WORKER_INTERACTION.md
+++ b/designs/DESIGN_AI_WORKER_INTERACTION.md
@@ -106,13 +106,30 @@ The parent processes the request (handling authentication, networking, and quota
 If the parent fails to execute the request (e.g., network failure after retries, invalid request), it returns an error.
 
 *   **Direction**: Parent -> Child
-*   **Format**:
+*   **New Format**:
+    ```json
+    {
+      "type": "error",
+      "payload": {
+        "message": "Rate limit exceeded, retry after 60s",
+        "class": "rate_limit",
+        "retry_after_secs": 60
+      }
+    }
+    ```
+*   **Error Classes**:
+    *   `fatal`: Permanent failure. `retry_after_secs` is omitted.
+    *   `rate_limit`: Provider quota or rate-limit failure. `retry_after_secs` is required.
+    *   `transient`: Retryable provider or transport failure. `retry_after_secs` is required.
+*   **Receive Semantics**: Object payloads are decoded into a typed `RemoteAiError`, preserving the message and error class for callers that need structured handling.
+*   **Legacy Format**:
     ```json
     {
       "type": "error",
       "payload": "Detailed error message string"
     }
     ```
+    String payloads remain accepted for backward compatibility and surface as a plain `Remote AI Error: ...` error.
 
 ### 5. Final Result (Worker -> Parent)
 Upon completion, the worker outputs the final review data. This message **does not** have a `type` wrapper, allowing it to be distinct from protocol messages, or identified by specific fields (e.g., `patchset_id`).
--- a/designs/DESIGN_AI_WORKER_INTERACTION.md
+++ b/designs/DESIGN_AI_WORKER_INTERACTION.md
@@ -31,7 +31,7 @@ sequenceDiagram
             P->>P: Handle Quota (Wait & Retry)
             P->>C: (Delayed Response)
         else Fatal Error
-            P->>C: {"type": "error", "payload": "Description"}
+            P->>C: {"type": "error", "payload": {"message": "Description", "class": "fatal"}}
         end
     end
 
@@ -106,7 +106,7 @@ The parent processes the request (handling authentication, networking, and quota
 If the parent fails to execute the request (e.g., network failure after retries, invalid request), it returns an error.
 
 *   **Direction**: Parent -> Child
-*   **New Format**:
+*   **Format**:
     ```json
     {
       "type": "error",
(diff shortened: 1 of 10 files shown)


derekbarbosa:
Hey there.

@rgushchin 
Running the `pr-review` skill resulted in the following:

```
✦ Here are the critical and medium concerns reformatted strictly according to the review standards:

 Safety / Bugs: In src/ai/mod.rs, classify_error_text checks for "429" anywhere in the error_text. This substring match is too aggressive. If an error string happens to contain the number 429 anywhere else
  (e.g., "failed to parse line 429"), it will be incorrectly classified as a RateLimit error and pause the entire system. We should pad the string check to ensure it matches the HTTP status code context.

    1 --- src/ai/mod.rs
    2 +++ src/ai/mod.rs
    3 @@ -251,7 +251,9 @@
    4      let lower = error_text.to_ascii_lowercase();
    5      if lower.contains("rate limit exceeded")
    6          || lower.contains("quota exceeded")
    7 -        || lower.contains("429")
    8 +        || lower.contains(" 429")
    9 +        || lower.contains(": 429")
   10 +        || lower.ends_with("429")
   11      {
   12          Some(AiErrorClass::RateLimit {
   13              retry_after: DEFAULT_RETRY_AFTER,

 Complexity / Logic / SOLID: In src/reviewer.rs, report_quota_error(retry_after) is called without bounding the duration. A maliciously formed or accidentally huge Retry-After header from an upstream proxy
  could cause the quota manager to block the global review process indefinitely. We should bound the maximum retry timeout to a reasonable upper limit (e.g., 1 hour) before passing it to quota_manager.

    1 --- src/reviewer.rs
    2 +++ src/reviewer.rs
    3 @@ -1679,8 +1679,10 @@
    4                                              Err(e) => {
    5                                                  match classify_ai_error(&e) {
    6                                                      AiErrorClass::RateLimit { retry_after } => {
    7 +                                                        let capped_retry =
    8 +                                                            retry_after.min(std::time::Duration::from_secs(3600));
    9                                                          quota_manager
   10 -                                                            .report_quota_error(retry_after)
   11 +                                                            .report_quota_error(capped_retry)
   12                                                              .await;
   13                                                          continue;
   14                                                      }
```

---

Looking at the diff, the addition of the thin trait is great
```
pub trait ClassifyAiError {
    fn ai_error_class(&self) -> AiErrorClass;
}

...

/// Classifies AI provider errors into retryable and non-retryable groups.
pub fn classify_ai_error(error: &anyhow::Error) -> AiErrorClass {
    if let Some(err) = error.downcast_ref::<openai::OpenAiCompatError>() {
        return err.ai_error_class();
    }

    if let Some(err) = error.downcast_ref::<gemini::GeminiError>() {
        return err.ai_error_class();
    }

    if let Some(err) = error.downcast_ref::<claude::ClaudeError>() {
        return err.ai_error_class();
    }

    if let Some(err) = error.downcast_ref::<crate::worker::prompts::ReviewError>() {
        return err.ai_error_class();
    }

    classify_error_text(&error.to_string()).unwrap_or(AiErrorClass::Fatal)
}
```

This enables some cleaner match blocks in reviewer.rs, allowing the implementation to be handled by the individual "provider" file (gemini.rs, claude.rs, etc) -- I like this approach 👍🏽 

```
         Err(e) => match classify_ai_error(&e) {
+                AiErrorClass::RateLimit { retry_after }
+                | AiErrorClass::Transient { retry_after } => {
+                    warn!(
+                        "API error ({}), pausing for {:?} before retry...",
+                        e, retry_after
+                    );
+                    tokio::time::sleep(retry_after).await;
+                }
+                AiErrorClass::Fatal => {
+                    return BenchmarkResult {
+                        commit: entry.commit,
+                        problem_description,
+                        found: false,
+                        status: "UNKNOWN".to_string(),
+                        explanation: format!("Evaluation failed: {}", e),
+                        findings_count,
+                        concerns_count,
+                        tokens_in,
+                        tokens_out,
+                        turns,
+                        duration_secs,
+                    };
+                }
+            },
```


I cross-compared my notes to that of the AI review bot, and it checks out to me.  I think an hour may be _too_ long of a timeout, but I am not sure what a "good one" would be for the errors you are encountering.


Please feel free to challenge me on these assumptions.

jgunthorpe:
It looks like both of the remarks are pre-existing, it just moved them around. I spent some time figuring out what is going on and the string processing looks to be some kind of hack just because the error gets marshaled through stdio as a simple string. To recover the lost semantic information it was using string parsing. It took some AI prodding but I got several commits that fix that root issue and pass the error class across the stdio with a json encoding so none of the semantic data is lost and then the sketchy string parsing can be entirely deleted. Also added a commit to cap the timeout just in case.

PR #174, opened 2026-05-11
ai: unify claude JSON instructions and error handling

### LINKED ISSUE #171: Vertex AI Claude provider drops response_format, Stage 1 always fails
Vertex AI Claude provider drops response_format, Stage 1 always fails with "failed to produce valid 'concerns' array"

## Environment

- sashiko: main branch (post PR #155 merge), also reproducible on v0.1.6
- Provider: vertex
- Model: claude-sonnet-4-6 and claude-opus-4-6 (both fail identically)
- Region: us-east5
- OS: Fedora 43
- Rust: 1.95.0

## Settings.toml

```toml
[ai]
provider = "vertex"
model = "claude-sonnet-4-6"
max_input_tokens = 180000

[ai.vertex]
project_id = "my-project"
region = "us-east5"
prompt_caching = true
max_tokens = 32000
thinking = "adaptive"
```

## What happens

`sashiko-cli local` extracts patches and applies them successfully, but every AI review fails at Stage 1:

```
[1/4] Extracting patches from <sha>... (1 commit)
[2/4] Starting review subprocess...
[2/4] Applying patches to worktree...
[3/4] AI review in progress...
Error: Stage 1 failed to produce valid 'concerns' array after 3 attempts — aborting review
```

`--no-ai` works fine (patch application succeeds).

The Vertex API endpoint itself works -- direct `rawPredict` calls with the same credentials return valid responses.

## Root cause

`src/worker/prompts.rs` Phase 0 (line 508) and Stage 1 both set `response_format: Some(AiResponseFormat::Json { schema })` to force structured JSON output.

However, `src/ai/vertex.rs` `generate_claude()` (line 255-262) constructs `VertexClaudeRequest` without forwarding `response_format`:

```rust
let vertex_req = VertexClaudeRequest {
    anthropic_version: "vertex-2023-10-16".to_string(),
    messages: claude_req.messages,
    max_tokens: claude_req.max_tokens,
    system: claude_req.system,
    tools: claude_req.tools,
    thinking: claude_req.thinking,
    // response_format is NOT forwarded
};
```

Claude's `rawPredict` endpoint does not support Gemini-style `response_format` / JSON schema natively. The `response_format` field is silently dropped, so Claude returns free-form text instead of the expected JSON structure. The JSON parser then fails.

## Suggested fix

Since Claude's API does not have a native `response_format` equivalent, the Vertex provider should convert `AiResponseFormat::Json { schema }` into a system prompt instruction. For example, append to the system message:

```
You MUST respond with ONLY a JSON object matching this schema: {schema}
No other text before or after the JSON.
```

This is how the `claude-cli` provider handles it (via Claude Code's built-in JSON mode).

Alternatively, the Vertex provider could use Claude's tool_use with a single tool whose input_schema matches the desired output schema, then extract the tool call result as the JSON response.

## Related issues

- #144 (Claude `thinking: {}` bug -- fixed by PR #155, but this issue persists)
- #107 (Claude `input_schema.type` error -- related but distinct)

derekbarbosa:
Centralizes JSON instruction logic in src/ai/mod.rs by adding AiResponseFormat::format_json_schema_instruction().

Improves claude-cli in src/ai/claude_cli.rs:
- Introduces ClaudeCliError implementing ClassifyAiError for better retry handling.
- Switches build_prompt to use the new centralized JSON instructions.

Unifies Claude API in src/ai/claude.rs:
- Injects the centralized JSON instructions into the system prompt.
- Removes redundant/duplicate formatting logic.

Updates error classification in src/ai/mod.rs to support the new ClaudeCliError.

fixes #171

### DIFF
--- a/src/ai/bedrock.rs
+++ b/src/ai/bedrock.rs
@@ -174,18 +174,57 @@ fn translate_request(
     thinking: Option<&str>,
     effort: Option<&str>,
 ) -> Result<ConverseParams> {
-    let system = request.system.as_ref().map(|s| {
-        let mut blocks = vec![SystemContentBlock::Text(s.clone())];
-        if enable_caching {
-            blocks.push(SystemContentBlock::CachePoint(
-                CachePointBlock::builder()
-                    .r#type(CachePointType::Default)
-                    .build()
-                    .expect("CachePointBlock build"),
-            ));
+    let mut system_blocks = Vec::new();
+
+    if let Some(s) = &request.system {
+        system_blocks.push(SystemContentBlock::Text(s.clone()));
+    }
+
+    // Collect any system messages from the messages array
+    for msg in &request.messages {
+        if let (AiRole::System, Some(content)) = (&msg.role, &msg.content) {
+            system_blocks.push(SystemContentBlock::Text(content.clone()));
         }
-        blocks
-    });
+    }
+
+    // Handle response format by appending instructions to the system prompt
+    // since Bedrock doesn't support it natively for all models.
+    if let Some(format) = &request.response_format {
+        match format {
+            crate::ai::AiResponseFormat::Json { schema } => {
+                let instruction = if let Some(schema) = schema {
+                    format!(
+                        "\n\nYou MUST respond with ONLY a JSON object matching this schema: {}\nNo other text before or after the JSON.",
+                        serde_json::to_string(schema).unwrap_or_default()
+                    )
+                } else {
+                    "\n\nYou MUST respond with ONLY a JSON object.\nNo other text before or after the JSON.".to_string()
+                };
+
+                if let Some(SystemContentBlock::Text(text)) = system_blocks.last_mut() {
+                    text.push_str(&instruction);
+                } else {
+                    system_blocks.push(SystemContentBlock::Text(instruction.trim().to_string()));
+                }
+            }
+            crate::ai::AiResponseFormat::Text => {}
+        }
+    }
+
+    if enable_caching && !system_blocks.is_empty() {
+        system_blocks.push(SystemContentBlock::CachePoint(
+            CachePointBlock::builder()
+                .r#type(CachePointType::Default)
+                .build()
+                .expect("CachePointBlock build"),
+        ));
+    }
+
+    let system = if system_blocks.is_empty() {
+        None
+    } else {
+        Some(system_blocks)
+    };
 
     let mut messages: Vec<Message> = Vec::new();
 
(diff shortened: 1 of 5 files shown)


HouMinXi:
Tested PR #174 on Vertex AI with `claude-sonnet-4-6`. Still fails at Stage 1:

```
Error: Stage 1 failed to produce valid 'concerns' array after 3 attempts
```

The `response_format` injection into system prompt looks correct (Phase 0 should work now), but Stage 1 uses **tool use**, and the tool schemas in `src/worker/prompts.rs` still have uppercase type names:

```rust
// prompts.rs:486-490
"type": "OBJECT",
"type": "ARRAY",
"type": "STRING"
```

Claude API requires lowercase (`"object"`, `"array"`, `"string"`). Gemini accepts both cases but Claude rejects uppercase with:
```
tools.0.custom.input_schema.type: Input should be 'object'
```

This is the same issue as #107. PR #108 (closed) fixed both problems (response_format + uppercase schemas), but only the response_format half made it into this PR.

Quick fix -- in `src/worker/prompts.rs`, change all inline JSON schema types to lowercase:
```diff
-"type": "OBJECT"
+"type": "object"
-"type": "ARRAY"
+"type": "array"
-"type": "STRING"
+"type": "string"
```

Also check `src/worker/tools.rs` if it has the same pattern (I didn't find uppercase there, but worth double-checking).

derekbarbosa:
Reworded some commits, added some more error handling for claude_cli, and unified JSON prompt building/formatting for claude responses.

derekbarbosa:
Addressed some `review-pr` concerns, mainly:

```
src/ai/claude.rs:L322-326: 🟡 risk: system_blocks.push appends a new text block. Some Anthropic-compatible proxies (e.g. AWS Bedrock, Google Vertex AI) might reject multiple consecutive
text blocks in the system prompt array. Consider appending the instruction to the last text block if one exists.

src/ai/mod.rs:L95: 🔵 nit: if serde_json::to_string fails (unwrap_or_default returns ""), the schema instruction becomes invalid JSON matching this schema: .. Since schema is dynamic,
might be safer to handle serialization errors gracefully (e.g., fallback to schema-less instruction).
```

Tested using us-east5 endpoint, E2E patch submission successful with sonnet & opus models via vertex.

ready for review! @rgushchin

derekbarbosa:
Hi @rgushchin, sorry for the noise. Seems like there are others encountering the JSON encoding issue. Kindly asking for a review :-)

anders-heimer:
Hi Derek,

Thanks for the PR. I ran a check against the JSON-format failures collected from my various benchmark runs.  The response-format side looks useful and centralizing the JSON response-format instruction should help cases where the model returns prose or JSON that misses the required top-level schema fields.

 I also replayed a small matrix of observed JSON/format failure classes. The PR helps with the instruction/schema side.

Br, Anders

derekbarbosa:
Hey Anders,

Thank you for posting your findings, it certainly helps. I'm glad the proposed changes address some of the formatting issues. Hopefully pairing this with TodoWrite may increase the consistency wrt failures across the board

derekbarbosa:
Hi @rgushchin -- anything else you'd like me to test here, so that we can close the original issue?

derekbarbosa:
rebased on latest `main`

rgushchin:
Merged, thanks

PR #177, opened 2026-05-11
Embargo updates

kuba-moo:
Small and sneaky update for the embargo handling.

**Small**. First, update the formatting of the embargo message a little. People are spooked out by it, thinking that Sashiko found some major security issue. So spell out that the embargo is due to mailing list policy. Second, add a bypass token. Not much UI handling at this point, just let user pass the magic token in to reveal the review. This is occasionally useful if we need to apply something ASAP and want to check that Sashiko didn't discover any issues (hence the somewhat intentional lack of smooth UI integration, this should not be an easy to use feature).

**Sneaky**. The token embargo bypass is useful for another reason, and this is something for a discussion more than directly related to this PR. I was wondering how we would tackle the multi-model reviews. Codex is coming to Bedrock, so for networking we have a choice of either adding multi-model support to Sashiko within one instance, and then everyone runs all the models they can. Or, option 2, we keep sashiko.dev as the "main" interface to Sashiko and "federate" the reviews. In the first option networking sashiko would run Claude+Codex and main sashiko would run Gemini. In the federated model all instances do their reviews (up to and excluding the dedup stage), then they wait and gather raw comments from all other connected instances, and then they dedup / false positive filter comments from all instances. I like the federated model not just because I'm cheap but also it shows user all findings in one interface and one email. Coming back to this PR - for federated model we want to let the instances read the reviews before the embargo expires tho, otherwise we're injecting 24h of latency before the dedup / false positive pass.


rgushchin:
Thanks, Jakub!

Message wise - good call, I've been already asked few times what embargo means, happy to merge immediately.

Re tokens - I'm a couple of days away from merging a big PR which will enable the authorization for maintainers. I'd prefer to build something like this on top, e.g. maybe just allow maintainers to see all and manually cancel embargoes.

Also agree on federated model and keeping sashiko.dev as the main interface. I'm building it for preexisting bugs so that various tools can discover bugs in the existing code, generate reproducers and potentially fixes and feed it into sashiko.dev, which will work as the interface for maintainers. I think federated reviews fit nicely into this model.

kuba-moo:
Sounds good! Can you cherry pick just the text changes or should I re-submit the PR with just the one commit?
Happy to wait for your auth work instead of the hacky tokens.

rgushchin:
Sure, will cherry-pick it and close the PR. Thanks!

Issue #128, opened 2026-04-24
feature: add a "mocked" database file for development purposes


derekbarbosa:
currently, it can be difficult to test functional changes to sashiko (UI, processing, etc).
a sample, dummy database would allow for quicker visual verification of changes.

a database file located in scripts/ or tests/ (or even in the root of the repository) labeled `sashiko-mock.db` with "realistic" mocked stats would be easier to test against, and may even allow us to perform some better integration testing in the future.

derekbarbosa:
thinking that potentially JSON fixtures akin to the benchmarks file may be a better approach :)

rgushchin:
I can probably run a tiny benchmark and commit the resulting database.

TheSven73:
Using the prompt below with antigravity + Gemini 3.5 Flash High resulted in a clean, workable, single-page `src/bin/prime_db` which did a believable job of populating the database. Two or three extra turns will probably give you whatever fancy additions you want :)

> The production sashiko.db contains 26,000 patchsets and 26,000 messages. create a Rust executable called `prime_db` in src/bin which creates
  a fresh sashiko.db with that amount of patchsets and messages. The text should consist of random words, yet plausible-looking, i.e. the
  subjects should start with [PATCH vX Y/Z] where X, Y, Z are numbers. 90% of patchsets should be "Reviewed" with no comments. the remaining 10%
  should be "Reviewed" with one or two findings, consisting of randomly generated words. Prefer simplicity. Reuse db schema and code from
  sashiko.


> now improve `prime_db` by adding review comments to the database. each patchset with findings should also have review comments. Create these
  with random words. One or two sentences per finding.

derekbarbosa:
@TheSven73 thank you for this. I played around with a scratch branch previously to create some JSON fixtures to generate a dummy DB -- but I have good success populating a DB with `--no-ai --download $NUM` 

granted, it doesn't do much to show you real "review stats", but I have found it helpful for testing UI changes and git_ops/ingestor testing.

PR #187, opened 2026-05-12
feature: allow for agnostic git backends

derekbarbosa:
Multi-Forge Integration via ForgeProvider Trait

Patches are currently ingested exclusively via NNTP (mailing list).
This patch adds a ForgeProvider trait that abstracts webhook-based
ingestion from code forges, with built-in implementations for GitHub and
GitLab:

- ForgeProvider trait: validate_event(), parse_payload() — any forge
- that speaks webhooks can implement this interface.
- ForgeRegistry: runtime registry of providers, pre-loaded with GitHub
- and GitLab. Third-party providers can be registered at startup.
- ForgeMetadata: normalized struct (repo_url, base_sha, head_sha,
- pr_number, pr_title, pr_url) — backend-agnostic representation of
- a PR/MR event.
- ForgeSettings in Settings.toml: enabled, provider, webhook_secret,
- api_token.
- Webhook endpoint (/api/webhook/:provider): receives forge events, validates
- via the registered provider, extracts commit range, and feeds patches into
- the existing review pipeline.
- DB schema additions: mr_url, mr_title, mr_number, slug columns on
- patchsets table for forge metadata tracking.

The forge layer is entirely opt-in. When [forge] enabled = false (default),
no webhook routes are registered and the NNTP pipeline is unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,7 +28,7 @@ dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
- "zerocopy 0.8.46",
+ "zerocopy 0.8.52",
 ]
 
 [[package]]
@@ -158,9 +158,9 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "autocfg"
-version = "1.5.0"
+version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
+checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
 
 [[package]]
 name = "aws-config"
@@ -183,7 +183,7 @@ dependencies = [
  "bytes",
  "fastrand",
  "hex",
- "http 1.4.0",
+ "http 1.4.2",
  "ring",
  "time",
  "tokio",
@@ -206,9 +206,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.17.0"
+version = "1.17.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ec2f1fc3ec205783a5da9a7e6c1509cc69dedf09a1949e412c1e18469326d00"
+checksum = "4342d8937fc7e5dd9b1c60292261c0670c882a2cd1719cfc11b1af41731e32ad"
 dependencies = [
  "aws-lc-sys",
  "untrusted 0.7.1",
@@ -217,14 +217,15 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-sys"
-version = "0.41.0"
+version = "0.42.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1a2f9779ce85b93ab6170dd940ad0169b5766ff848247aff13bb788b832fe3f4"
+checksum = "6d9ceb1da931507a12f4fccea479dccd00da1943e1b4ae72d8e502d707361444"
 dependencies = [
  "cc",
  "cmake",
  "dunce",
  "fs_extra",
+ "pkg-config",
 ]
 
 [[package]]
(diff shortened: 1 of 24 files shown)


rgushchin:
Hey! I asked Gemini to review this PR and it seems to be very concerned :) It makes me too. Db safety is top priority. Please, take a look. Overall stability > velocity at this point, Sashiko is actively used by the kernel community and just keep it afloat is p0.

🚨 Critical Regression Risks

  1. Database Migration Crash on Startup (Existing Installs)
  There is a major issue with how the new slug column is handled in existing databases:
   * In src/schema.sql, the PR adds CREATE UNIQUE INDEX IF NOT EXISTS idx_patchsets_slug ON patchsets(slug) WHERE slug IS NOT NULL;.
   * In src/db.rs, Database::migrate() executes schema.sql before it runs manual migrations (try_add_column).
   * Because of this order, existing databases will crash on startup: SQLite will attempt to create an index on the slug column before it exists, throwing a no such column: slug error.
   * Furthermore, the PR completely missed adding self.try_add_column("patchsets", "slug", "TEXT").await; to db.rs. So even if schema.sql didn't crash, subsequent SELECT queries that now
     expect the slug column will fail.

  2. Mutual Exclusion of Ingestors (Loss of Functionality)
  In src/main.rs, the PR modifies the startup logic to disable the Lore/NNTP ingestor completely if forge integration is enabled:

   1 let ingestor_handle = if !settings.forge.enabled {
   2     // Starts Lore/NNTP Ingestor
   3 } else {
   4     info!("Forge integration is enabled. Lore/NNTP ingestor is disabled.");
   5     // Spawns a pending future
   6 };
  This means a user cannot simultaneously monitor mailing lists and accept GitHub/GitLab webhooks. If a user enables forge features, their mailing list integration will silently stop
  working.

  ⚠️ Moderate Risks & Behavior Changes

  1. Subsystem Mapping Fallback Logic
  In src/lib.rs -> identify_subsystems(), the logic was updated to use custom mapping rules from Settings.toml. However, the code uses:

   1 if !matched && mapping.is_empty() {
   2     // Fallback to default linux-kernel, netdev, etc. checks
   3 }
  If a user specifies even one custom subsystem mapping in their configuration, mapping.is_empty() evaluates to false, and the default fallback logic is bypassed completely. Users adding a
  custom rule might be surprised when Sashiko stops identifying standard lists like LKML.

derekbarbosa:
Hey!

> Because of this order, existing databases will crash on startup: SQLite will attempt to create an index on the slug column before it exists, throwing a no such column: slug error.
Furthermore, the PR completely missed adding self.try_add_column("patchsets", "slug", "TEXT").await; to db.rs. So even if schema.sql didn't crash, subsequent SELECT queries that now
expect the slug column will fail.

Interesting that the `pr-review` didn't catch this bug in the first few iterations of the RFC! I will run some reviews of my own and attempt to test this behavior.

Would you mind following up on #128? It would allow me to test this with a bit more certainty 😅

Otherwise I can just implement the proposed fixes and re-run the reviews.

> Mutual Exclusion of Ingestors (Loss of Functionality)
> In src/main.rs, the PR modifies the startup logic to disable the Lore/NNTP ingestor completely if forge integration is enabled <snip/>
> This means a user cannot simultaneously monitor mailing lists and accept GitHub/GitLab webhooks. If a user enables forge features, their mailing list integration will silently stop
> working.

Is this actually critical? I would imagine a user spinning up a sashiko instance to track a github repository isn't particularly interested in also tracking NNTP. If it is to be desired though, it is an easy enough fix.

 > If a user specifies even one custom subsystem mapping in their configuration, mapping.is_empty() evaluates to false, and the default fallback logic is bypassed completely. Users adding a
custom rule might be surprised when Sashiko stops identifying standard lists like LKML.

At first, this was intended behavior to prevent any collisions from happening with custom subsystem mappings. I realize now that it needs to be reworked a bit.


Thanks for the review! v2 coming soon :-)

derekbarbosa:
V1->V2:

- Addressed review comments
- Additional tests around `nntp` disable/enable, custom subsystem identification, forge settings, etc
- Refactored `regex::Regex::new` compilation for performance.
- Replaced `.unwrap()` calls with `if_let` patterns in the mr_title extraction path for git forge
- Rebased onto latest main.

derekbarbosa:
v2 -> v3:

- rebased onto v0.2

derekbarbosa:
V3 -> V4: 
- rebased on latest `main`
- added new documentation disclaimer

PR #134, opened 2026-04-28
Add sashiko-review for serverless reviews

ctiller:
Introduce the sashiko-review-now command-line tool, enabling patch evaluations in a standalone environment independent of full server infrastructure.

Update prompt loading mechanisms to support embedded fallbacks, which removes the hard dependency on prompts directory availability and simplifies distribution.

Improve tracing clarity by adding generation request indicators and tool return previews to the multi-stage reviewer logging.

Adjust PromptRegistry tests to accept optional directory parameters.

PR #188, opened 2026-05-12
feature: enable custom prompts/review-stages

derekbarbosa:
Sashiko's review pipeline currently has a fixed structure: hardcoded stage
instructions, a static set of AI tools, and no mechanism for users to inject
domain-specific review guidance.

Those adopting sashiko for their own
kernel trees or non-kernel projects must either:

- Fork the repository and modify prompt files directly, creating a maintenance burden when upstream evolves.
- Accept the default review pipeline as-is, missing project-specific review criteria and tooling.

Neither option scales. Someone reviewing storage drivers needs different
guidance someone reviewing networking code, but today both run the
identical 9-stage pipeline with identical prompts.

This patch introduces a configuration-driven customization layer that sits
between the user and the review pipeline. All changes are opt-in; the default
behavior is completely unchanged when no configuration is provided.

_The following features are enabled by this PR:_

**Custom Prompts Directory**

Users can point sashiko at their own prompt files via Settings.toml or the
--prompts CLI flag. The directory can be a local path, a Git repository URL
(cloned and cached locally), or an HTTP(S) URL. The default
third_party/prompts/kernel directory continues to work unchanged.

**Stage Configuration**

A new stages.toml file controls the review pipeline:

- Enable or disable individual stages
- Override the instruction file for any stage
- Specify per-stage supporting documents
- Add entirely new custom stages beyond the default 9

This allows teams to skip irrelevant stages (e.g., disable the hardware review
stage for a pure-software project) or add domain-specific stages (e.g., a
compliance analysis stage).

**Template Variables**

Prompt files can contain {{variable}} placeholders that are substituted at
load time from configuration. Built-in variables include {{date}} and
{{year}}. Users define custom variables in [prompts.variables] to inject
organization names, target architectures, or other project-specific context
into prompts without modifying the prompt files themselves.

**Tool Filtering**

Users can allowlist or denylist which built-in AI tools are available during
review. This is useful for restricting the AI to read-only operations, or
limiting tool access in constrained environments.

**Custom Tool Definitions**

Users can register shell-based tools that the AI can invoke during review.
Custom tools are defined in Settings.toml with a JSON schema for parameters,
a shell command template, and an optional path allowlist for security. This
enables integration with project-specific linters, static analyzers, or other
external tools without modifying sashiko's source.

This PR:

- does not change the default review behavior. Without a [prompts] or
- [tools] section in Settings.toml, sashiko operates identically to before.
- does not add new AI providers or modify the AI interaction protocol.
- does not change the review output format.


derekbarbosa:
V1-> V2:

- Corrected the environment variable naming convention in README.md to SASHIKO__AI__PROVIDER (double underscore after the prefix) to align with the config-rs
implementation.

- Replaced hard panics (.expect()) in src/bin/review.rs with proper error propagation using anyhow::Context. This prevents the application from crashing abruptly
if AI provider initialization or prompt loading fails.

- Unit test in src/worker/prompts.rs to codify the backward compatibility of stage filtering. This ensures that if a stage is omitted from a custom stages.toml, it correctly defaults to "enabled," preserving the expected behavior for existing configurations.

- Rebased onto main

floatious:
I like your change, as I think it is much nicer to have each stage defined in a separate .md file, rather than hardcoded in the source.

Question: If we compare with the /kseries skill in review-prompts:
https://github.com/masoncl/review-prompts/blob/main/kernel/skills/kernel.md
https://github.com/masoncl/review-prompts/blob/main/kernel/review-core.md

Would it be possible to create a similar skill for Sashiko, than could be imported directly in the agent and executed, to perform all the review stages in order, or would this not be possible because of the way Sashiko transitions between stages?

Since is seems that skills are more or less compatible between agents, having a /sashiko-kseries skill would basically make a serverless standalone binary as proposed in: https://github.com/sashiko-dev/sashiko/pull/134 unnecessary.

derekbarbosa:
Hi @floatious, sorry for the delay in reply.

Sure, I can push up an experimental branch. 

> Would it be possible to create a similar skill for Sashiko, than could be imported directly in the agent and executed, to perform all the review stages in order, or would this not be possible because of the way Sashiko transitions between stages?

Maybe. I would imagine that in a long-running session, you'd get issues with some models (Claude, Gemini) compacting the "conversation context" -- potentially missing out findings.

However, I don't see why that can't be circumvented by writing findings out to a file or something between stages, propagating them forward in a similar fashion.

I'll hack something together up towards the end of the weekend/early-next-week, and push up a branch. Does that sounds good?

In the meantime, would you mind filing an issue with a small description, just to keep it separate from the RFC? :)

floatious:
> I'll hack something together up towards the end of the weekend/early-next-week, and push up a branch. Does that sounds good?
> 

No stress at all, I just thought that having the review implemented as a skill might be useful, as it makes it super easy for people to get the "Sashiko review" locally, without the need for a Sashiko server.


> In the meantime, would you mind filing an issue with a small description, just to keep it separate from the RFC? :)

Here you go:
https://github.com/sashiko-dev/sashiko/issues/219


Thank you!

derekbarbosa:
V2-> V3:

Rebased onto 0.2

derekbarbosa:
v2 -> v3:
- rebased on latest main
- added documentation disclaimer

designs/DESIGN_CUSTOM_PROMPTS_AND_TOOLS.md
```diff
@@ -0,0 +1,442 @@
...
+
+- **Local path**: Absolute or relative to the working directory.
+  ```toml
+  [prompts]
+  directory = "/opt/my-prompts"
+  # or
+  directory = "./custom-prompts"
+  ```
+
+- **Git URL**: Cloned to `.sashiko-cache/prompts/{md5(url)}`.
```
chrissicool: This prompts directory download URL hash is the only reason for a new dependency on the md5 crate. Sashiko depends on sha2 already for interaction with Git. Would you mind switching to sha256?

derekbarbosa: hey, thank you for the review. sure, I missed this in my latest rebase.

derekbarbosa:
v3 -> v4:

- rebased on latest main
- dropped md5 dependency
- edited docs

derekbarbosa:
after some offline discussion with @rgushchin -- closing this PR for now. will revisit.

PR #197, opened 2026-05-15
prompts: I3C_OR_I2C is a valid Kconfig symbol.

groeck:
Sashiko complained:

> +	depends on I3C_OR_I2C

Is this a valid Kconfig symbol? It looks like it will evaluate as a single config symbol named CONFIG_I3C_OR_I2C instead of a logical OR. Could this cause the driver to be silently unbuildable? Perhaps it should be written as depends on I3C || I2C?

Tell it that it is a valid symbol.

### DIFF
--- a/third_party/prompts/kernel/subsystem/kconfig.md
+++ b/third_party/prompts/kernel/subsystem/kconfig.md
@@ -91,6 +91,10 @@ approaches:
   depends on SOME_SUBSYSTEM
   ```
 
+## Unusual but valid Kconfig symbols
+
+- `I3C_OR_I2C` is a valid Kconfig symbol.
+
 ## Quick Checks
 
 - **Selected symbol existence**: Verify every `select FOO` references a


derekbarbosa:
The `prompts`  directory is occasionally sync'd to Chris Mason's [review-prompts ](
https://github.com/masoncl/review-prompts) repository.  This change seems like one that would benefit other users of those prompts as well, as I'm sure other models could get confused with the symbol under certain conditions.

Would it be better suited [there](https://github.com/masoncl/review-prompts/blob/main/kernel/subsystem/kconfig.md)?

rgushchin:
@derekbarbosa this repo already contains hwmon-specific prompts, which are not synchronized to Chris’s repo. I’m going to merge this one too. We gonna sync it all a bit later.

derekbarbosa:
You're right. I forgot about those. 

Yes. We certainly can do "batches" of syncs every now and then! :)

PR #199, opened 2026-05-16
Drop crap emails to krzk

krzk:
Amount of noise, incorrect comments and misleading statements you produce, plus recently introduced fake reviewer's statement of oversight, is just too much.


rgushchin:
Krzysztof, your email is dropped from the configuration by commit 7ad4aea47a09f6263a6da8102ab9acd3b5c699bd (rollout in progress). The aggressive language of your commit prevents me from merging it directly. Thanks.

PR #213, opened 2026-05-23
review: expose failed review diagnostics in raw logs

smcdef:
Review failures were stored as review-level result descriptions, while the patchset page mostly surfaced patchset.failed_reason. When the review worker failed before producing a final JSON payload, the raw log page could also be empty because no interaction logs had been persisted.

Keep patchset details concise and route failure diagnosis through the existing raw-log link. Make the raw log page show the final failure reason above the interaction log, then fall back to saved raw output or input context when no structured interaction log was recorded.

Capture partial stdout, stderr, and AI turn summaries while the review tool runs, keep those logs if completion later stores no log payload, and store process output with explicit process-log metadata so it can be rendered separately from AI interactions.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -762,7 +762,7 @@ impl Database {
     ) -> Result<()> {
         self.conn
             .execute(
-                "UPDATE reviews SET status = ?, result_description = ?, summary = ?, interaction_id = ?, inline_review = ?, logs = ? WHERE id = ?",
+                "UPDATE reviews SET status = ?, result_description = ?, summary = ?, interaction_id = ?, inline_review = ?, logs = COALESCE(?, logs) WHERE id = ?",
                 libsql::params![status, result, summary, interaction_id, inline_review, logs, review_id],
             )
             .await?;
--- a/src/reviewer.rs
+++ b/src/reviewer.rs
@@ -38,6 +38,9 @@ use tokio::process::Command;
 use tokio::sync::Semaphore;
 use tracing::{error, info, warn};
 
+const REVIEW_LOG_MAX_ENTRIES: usize = 2000;
+const REVIEW_LOG_STRING_LIMIT: usize = 12_000;
+
 #[derive(Clone)]
 struct ReviewContext {
     semaphore: Arc<Semaphore>,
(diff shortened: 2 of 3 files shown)


smcdef:
After apply the patch, developers will know the reason about the failure.

image

rgushchin:
Thanks for the PR!
I like the idea of exposing LLM logs for a failed review, I don't want to expose process logs through the frontend (maybe later, Idk).

smcdef:
> Thanks for the PR! I like the idea of exposing LLM logs for a failed review, I don't want to expose process logs through the frontend (maybe later, Idk).

OK, only expose failure reason. I've updated a new version.

image

derekbarbosa:
Hi -- running a `/review-pr`

```
🔴 Bug / Logic (Failed AI Requests Loop):
At L1732, the PR captures outgoing AI requests:

  push_review_log_entry(&mut captured_log, request_entry);
  At L1809, it captures responses:

  push_review_log_entry(&mut captured_log, response_entry);
  However, if the API call fails (e.g. rate limit, connection error), it captures the error:

   1 push_review_log_entry(
   2     &mut captured_log,
   3     json!({
   4         "role": "system",
   5         "content": format!("AI provider error: {message}")
   6     }),
   7 );

If an API fails in a loop, it pushes the request, then the error. Because REVIEW_LOG_MAX_ENTRIES is 2000, a rapid failure loop will quickly fill the 2000 entry limit with error logs. 

The truncation happens on push_review_log_entry, meaning the first 2000 entries are kept, and the last (which contains the fatal error that broke the tool) might be truncated out entirely, defeating the diagnostic purpose of this PR.
```


I took a closer look independent of the LLM:
```
   1   if log.len() < REVIEW_LOG_MAX_ENTRIES {
   2       log.push(entry);
   3   }
```
If the AI worker loops 1,000 times on a rate limit, the limit is hit.  Subsequent,  useful errors or final states will just hit the "Additional review log entries were omitted" fallback and be lost.


Continuing with the review skill:

 ```
🟡 Style / Nit (Unnecessary Option Allocation):
  In src/db.rs:L762:

"UPDATE reviews SET status = ?, result_description = ?, summary = ?, interaction_id = ?, inline_review = ?, logs = COALESCE(?, logs) WHERE id = ?",

logs is passed as Option<&str>. If the logs column in the database is large (which it is, up to 12k chars per entry * 2000 entries = 24MB), running COALESCE on a large text column for every UPDATE is unnecessary database overhead. Since the rust layer knows if logs is Some or None, it
should conditionally generate the SQL query rather than relying on COALESCE.

🔵 Style / Nits / DRY (src/reviewer.rs:L1523) - Double Serialization
The PR does serde_json::to_string_pretty and then immediately serde_json::from_str::<Value>. 
It is highly inefficient to serialize and deserialize the entire log array just to clone it into another JSON tree. It should just clone() the Vec<Value> and insert it directly as a Value::Array.
```

PR #216, opened 2026-05-24
review: preserve dismissed concerns to reduce false positives

smcdef:
## Summary

This PR improves the multi-stage review flow by preserving disproved candidate concerns and using them to reduce false positives before final verification.

Changes include:

  - Require stages 1-7 to return both `concerns` and `dismissed_concerns`.
  - Aggregate and normalize dismissed concerns alongside reported concerns, including stage source information.
  - Extend Stage 8 deduplication to consolidate both concerns and dismissed concerns.
  - Add a new Stage 9 to resolve conflicts between consolidated concerns and consolidated dismissed concerns.
    - Both sides are treated as untrusted hypotheses.
    - The stage verifies the actual code before deciding whether to keep or discard a concern.
    - Concerns are discarded only when dismissed-concern evidence concretely disproves the same root cause, code path, or failure mode.
  - Shift verification/severity estimation to Stage 10 and LKML report generation to Stage 11.

## Motivation

I found that some review comments for certain commits were false positives. After checking the raw review logs, I noticed that a few stages could incorrectly report the same candidate issue as a concern, while another stage might correctly investigate and dismiss that concern.

However, if that stage did not report any actual concerns, its dismissal evidence was discarded by the old pipeline. As a result, the final verification stage did not receive this important information and could fail to suppress the incorrect concern, causing a false-positive finding to be reported.

This PR preserves `dismissed_concerns` from stages 1-7 and feeds them into the later consolidation and conflict-resolution flow. In local experiments, this helps the LLM become more cautious when it sees conflicting conclusions about the same root cause, code path, or failure mode, and encourages it to re-check the actual code before reporting the issue.

## Testing

Tested locally with commits that previously triggered false-positive review comments. With `dismissed_concerns` propagated through the pipeline and reconciled before final verification, those false positives were resolved with high probability.


rgushchin:
JFYI: it looks like this change regressed the benchmark a bit. I still like the direction and the regression is manageable, so I'm not going to revert it and hopefully will fix it with some new changes.

rgushchin:
But realistically such changes should go through the full benchmark.

smcdef:
Hi Roman,

I’m sorry to hear that there's a regression. I previously only ran the simplest benchmark test (benchmark_tiny.json) and didn't notice any issues back then. Because I didn't have access to more tokens for testing, I wasn't able to run the broader benchmarks. On the bright side, I did test some previous commits that frequently triggered false positives and found that this approach successfully reduced them, so I still feel it's a step in the right direction.

That said, I'm really glad to hear that the regression is manageable! If you're open to sharing, I'd love to learn more about the issues you ran into.

PR #217, opened 2026-05-24
ai: contain Kiro ACP stream failures

anders-heimer:
## Summary

Kiro ACP `session/prompt` is a long-lived stream. When that stream fails,
Sashiko currently retries the same expensive turn without enough
provider-specific containment.

This PR adds Kiro-specific handling for those failures:

- classify Kiro ACP stream, rate-limit, provider, and permanent errors
- include bounded, redacted ACP diagnostics
- block retries after side-effect-looking ACP updates
- preserve terminal budget failures across review-bin stdio
- bound Kiro retry loops with per-turn budgets, wall-clock guards,
  idle watchdogs, and a process-local circuit breaker

## Patch Layout

1. `ai: improve kiro acp diagnostics`

   Capture malformed ACP stdout separately from stderr and surface
   redacted JSON-RPC `error.data`.

2. `ai: classify transient kiro acp errors`

   Classify known Kiro stream failures, throttling, provider availability
   failures, and permanent auth/configuration errors.

3. `ai: block kiro retries after side effects`

   Treat retryable-looking failures as fatal once ACP updates suggest a
   possible tool, command, or file mutation.

4. `ai: preserve terminal budget errors over stdio`

   Carry fail-fast budget errors through the review-bin AI stdio protocol
   so subprocess boundaries do not turn them into ordinary retryable
   remote errors.

5. `ai: contain kiro acp stream failures`

   Add Kiro-only retry budgets, same-error streak caps, turn wall-clock
   limits, ACP idle telemetry, prompt cancellation, request-id extraction,
   and a process-local circuit breaker.

## Notes

The only shared runtime plumbing is the review-bin terminal error marker.
It is generic, but included here because Kiro containment needs terminal
provider failures to survive stdio serialization.

This PR does not include follow-up handling for empty successful Kiro
responses, Kiro response parser quirks, or broader CLI diagnostic logging.


derekbarbosa:
Hi Anders, would you mind rebasing your changes?

anders-heimer:
Hi Derek, I have rebased the changes and run a few reviews before pushing. 

Thanks, Anders

derekbarbosa:
Hi Anders,

Thank you for the rebase. I am pasting the output of the review skill here:

```
🔴 1. terminal_kind is written but never read on the real stdio path
The worker tags terminal budget errors with terminal_kind (src/reviewer.rs:2026, 2054, 2125), but the only code that interprets it is decode_stdio_ai_response (src/ai/mod.rs:334), which is #[allow(dead_code)] and only exercised by unit tests.
The production reader is start_stdin_reader (src/ai/mod.rs:865-871), which deserializes RemoteAiErrorPayload and calls into_error():
pub fn into_error(self) -> RemoteAiError {
    RemoteAiError {
        message: self.message,
        class: self.class,
    }
}
terminal_kind is silently dropped. The dispatch signature is Result<AiResponse, RemoteAiError>, so a ReviewError::BudgetExceeded cannot survive the boundary. The "preserve terminal budget errors over stdio" commit has no production effect — the marker is dead data. (The class: Fatal fallback still fails fast, which may mask this, but the mechanism is incomplete.)
Fix — carry terminal_kind through so it isn't discarded:
--- a/src/ai/mod.rs
+++ b/src/ai/mod.rs
@@ -252,7 +252,8 @@ pub(crate) struct RemoteAiErrorPayload {
 pub struct RemoteAiError {
     pub message: String,
     pub class: AiErrorClass,
+    pub terminal_kind: Option<RemoteTerminalKind>,
 }
@@ -243,10 +244,11 @@ impl RemoteAiErrorPayload {
     pub fn into_error(self) -> RemoteAiError {
         RemoteAiError {
             message: self.message,
             class: self.class,
+            terminal_kind: self.terminal_kind,
         }
     }
 }
@@ -894,6 +896,7 @@ pub(crate) fn start_stdin_reader(registry: Arc<IpcRegistry>) -> tokio::task::Jo
             .abort_all(RemoteAiError {
                 message: "IPC channel disconnected (stdin closed)".to_string(),
                 class: AiErrorClass::Fatal,
+                terminal_kind: None,
             })
             .await;
The awaiting side of dispatch must then inspect terminal_kind (or re-derive ReviewError::BudgetExceeded) to actually fail fast. Alternatively, delete decode_stdio_ai_response and wire its terminal check directly into the "error" arm of start_stdin_reader.
🟠 2. Kiro turn timeout relaxed from 300s → 1200s
timeout_secs was settings.ai.api_timeout_secs (default 300s). It's now kiro_max_turn_wall_clock() = 20 * 60 (1200s), and api_timeout_secs no longer bounds Kiro turns at all (src/ai/mod.rs:591-599, src/ai/kiro_cli.rs:129, 286-292). This is documented as intentional, but a single turn now holds a worker/LLM semaphore permit for up to 4× longer. Confirm no downstream guard (review active-time, llm_semaphore, per-review deadline) assumed ≤300s.
🟠 3. initialize / session/new have no idle timeout
Only session/prompt passes idle_timeout: Some(...) (src/ai/kiro_cli.rs:1959-1963). initialize and session/new use RpcReadOptions::default() (idle_timeout None), so a Kiro process that hangs during startup (e.g., auth/network) is bounded only by the outer 1200s turn timeout. Combined with finding #6, a non-conforming line during these phases stalls the full 20 minutes. Consider applying the idle timeout to all three phases.
🟠 4. Stricter response matching can turn a malformed response into a 20-minute hang
Original matched msg.get("id") == target_id directly. The new is_acp_json_rpc_message requires jsonrpc == "2.0" and (method or (id + result/error)) before a line counts (src/ai/kiro_cli.rs:1029-1044). A Kiro line carrying id == target_id but missing jsonrpc, or a response with only "id" and no result/error, is now silently ignored and never advances the idle timer — for initialize/session/new that means blocking until the 1200s outer timeout. Low likelihood against a spec-compliant Kiro, but a robustness regression versus the old direct id match.
🟡 5. session/cancel is effectively dead code
On session/prompt failure the code writes session/cancel (src/ai/kiro_cli.rs:1967-1992) and then immediately return Err(e), which drops KiroChildGuard and kills the process before the cancel can take effect. The cancel write never accomplishes anything. Either remove it, or start_kill() only after giving Kiro a moment to process the cancel.
--- a/src/ai/kiro_cli.rs
+++ b/src/ai/kiro_cli.rs
@@ -1965,26 +1965,7 @@ impl KiroCliProvider {
         )
         .await
         {
-            let cancel_id = next_id + 1;
-            match write_rpc_checked(
-                &mut stdin,
-                cancel_id,
-                "session/cancel",
-                json!({ "sessionId": session_id }),
-                &stderr_preview,
-                &stdout_preview,
-            )
-            .await
-            {
-                Ok(()) => debug!(
-                    provider = "kiro-acp",
-                    session_prompt_id = next_id,
-                    cancel_id = cancel_id,
-                    "kiro-cli ACP session/cancel sent after session/prompt failure"
-                ),
-                Err(cancel_error) => debug!(
-                    provider = "kiro-acp",
-                    session_prompt_id = next_id,
-                    cancel_id = cancel_id,
-                    error = %cancel_error,
-                    "kiro-cli ACP session/cancel failed after session/prompt failure"
-                ),
-            }
             return Err(e);
         }
🔵 6. Duplicated idle-timeout error construction (DRY)
read_rpc_response builds the same KiroCliError message twice — once in the checked_sub-else branch and once in the map_err branch (src/ai/kiro_cli.rs:1064-1075 vs 1077-1088). Extract a helper:
--- a/src/ai/kiro_cli.rs
+++ b/src/ai/kiro_cli.rs
@@ -1061,13 +1061,7 @@ async fn read_rpc_response(
         let next_line = async { lines.next_line().await };
         let next_line = if let Some(idle_timeout) = options.idle_timeout {
             let elapsed_since_acp = last_acp_line_at.elapsed();
             let Some(remaining) = idle_timeout.checked_sub(elapsed_since_acp) else {
-                let side_effects_seen = side_effects_seen.load(Ordering::Relaxed);
-                return Err(KiroCliError::new(
-                    format!(
-                        "kiro-cli ACP idle timeout after {:.2}s waiting for response {} (side_effects_seen={})",
-                        idle_timeout.as_secs_f64(),
-                        target_id,
-                        side_effects_seen
-                    ),
-                    timeout_error_class(side_effects_seen),
-                )
-                .into());
+                return Err(kiro_idle_timeout_error(idle_timeout, target_id, side_effects_seen).into());
             };
             timeout(remaining, next_line).await.map_err(|_| {
-                let side_effects_seen = side_effects_seen.load(Ordering::Relaxed);
-                KiroCliError::new(
-                    format!(
-                        "kiro-cli ACP idle timeout after {:.2}s waiting for response {} (side_effects_seen={})",
-                        idle_timeout.as_secs_f64(),
-                        target_id,
-                        side_effects_seen
-                    ),
-                    timeout_error_class(side_effects_seen),
-                )
+                kiro_idle_timeout_error(idle_timeout, target_id, side_effects_seen)
             })?
         } else {
             next_line.await
         };
with a new helper:
fn kiro_idle_timeout_error(
    idle_timeout: Duration,
    target_id: u64,
    side_effects_seen: &AtomicBool,
) -> KiroCliError {
    let side_effects_seen = side_effects_seen.load(Ordering::Relaxed);
    KiroCliError::new(
        format!(
            "kiro-cli ACP idle timeout after {:.2}s waiting for response {} (side_effects_seen={})",
            idle_timeout.as_secs_f64(),
            target_id,
            side_effects_seen
        ),
        timeout_error_class(side_effects_seen),
    )
}
🟡 7. Quadratic scan in push_json_value_for_marker_scan
Each array/object iteration re-computes buffer.chars().count() (O(n) over the accumulated buffer), making the classification scan O(n²) for large nested error data (src/ai/kiro_cli.rs:720-755). Bounded by KIRO_MAX_CLASSIFICATION_TEXT_CHARS (64 KiB), so it's a worst-case CPU cost of ~2B char ops on a huge adversarial error payload rather than a crash. Consider tracking a running char count instead of calling chars().count() per element.
🔵 8. .expect() in KiroChildGuard::child_mut
child_mut uses .expect("kiro child already cleaned up") (src/ai/kiro_cli.rs:74). Currently statically safe (the child is only taken in kill_and_wait/Drop, and child_mut is never called after), but GEMINI.md asks for .expect() only when proven non-panicking and documented. It's a latent panic if call ordering changes; a debug_assert/documented invariant would suffice.
🟢 Design alignment
The rest of the change aligns well with designs/DESIGN_KIRO_CLI_PROVIDER.md: the provider state moved behind KiroCliProvider::new(...) with pub(crate) fields plus an Arc<KiroProviderState> (src/ai/mod.rs:591-599, src/ai/kiro_cli.rs:49-57), and the SASHIKO__AI__KIRO_* containment knobs are correctly documented in docs/llm-providers.md. The per-turn retry budget, same-failure streak, wall-clock guard, and process-local circuit breaker are a coherent, well-tested containment model — the main gap is that its terminal signal (finding #1) never reaches the coordinator.
Test coverage is strong (20+ new unit/integration tests including a Linux-only reaping test and fake-ACP end-to-end tests). I did not run make check-pr/cargo test (static review only); recommend running make check-pr before merge.
```

anders-heimer:
Hi Derek,

Thanks for the review comments. I have addressed them and pushed an updated patchset.

One of the fixes affects more than the kiro provider though. Once token-budget exhaustion is detected, the review now stops without entering the patch-level or whole-review retry paths. Previously, the terminal marker was lost across the stdio boundary, so the review could perform unnecessary (?) retries before eventually failing.

This behavior applies to all providers. I believe this is desirable because retrying cannot recover from an exhausted token budget, but it is broader behavioral change than I initially intended.

Thanks, Anders

PR #225, opened 2026-05-28
Add clone discipline to review-pr skill

TheSven73:
Integrate clone discipline guidelines and examples into the review-pr skill.

Establishing these checks directly in the review directives helps scrutinize future commits for redundant or excessive uses of clone. This is particularly useful in preventing unnecessary deep copies and cloning anti-patterns from being introduced into the codebase.

### DIFF
--- a/skills/review-pr/SKILL.md
+++ b/skills/review-pr/SKILL.md
@@ -20,6 +20,8 @@ Perform a deep, scrutinizing code review of the target code against the gathered
 - **Safety:** Are there any risky operations (e.g., unchecked unwraps, poor error handling, thread safety issues, SQL injection risks)?
 - **Idiomatic Conventions:** Does the code use the language idiomatic patterns and the project-specific coding standards defined in `GEMINI.md`?
 - **SOLID/DRY Principles:** Reference [references/principles.md](references/principles.md) and evaluate the code for adherence to SOLID and DRY principles within the Rust context.
+- **Clone Discipline:** If the code calls `.clone()` on expensive-to-clone structures, reference
+[references/clone-discipline.md](references/clone-discipline.md) and evaluate the code for adherence to clone discipline.
 - **Test Coverage:** Are unit or integration tests missing for new logic?
 
 ### 3. Reporting
--- a/skills/review-pr/references/clone-discipline.md
+++ b/skills/review-pr/references/clone-discipline.md
@@ -1,0 +1,140 @@
+# Clone Discipline
+
+## Directive
+
+Detect and challenge redundant and/or excessive uses of `.clone()`.
+
+## Definition
+
+The "Attack of the Clones" anti-pattern in Rust refers to the excessive and often unnecessary use of the
+`.clone()` method to bypass ownership and lifetime challenges. It occurs when developers choose to create
+deep copies of data as a "path of least resistance" to satisfy the borrow checker, rather than
+addressing the underlying structural or ownership issues in the code.
+
+## Anti Pattern Examples
+
+### BAD: Cloning a `Struct` when a `&Struct` will do
+
+```rs
+// BAD: prefer using `line: &str` which eliminates the .clone() below
+fn make_lipographic(banned: char, line: String) -> String {
+    line.as_str().chars().filter(|&c| c != banned).collect()
+}
+
+fn main() {
+    let passage = String::from("If Youth, throughout all history, had had a champion to stand up for it");
+    assert_eq!(make_lipographic('e', passage.clone()), passage);
+}
+```
+
+### BAD: Cloning a `Vec` when iterating
+
+```rs
+fn main() {
+    let mut bad_letters = vec!['e', 't', 'o', 'i'];
+    // BAD: prefer `for l in &bad_letters {`
+    for l in bad_letters.clone() {
+        // do something here
+    }
+    bad_letters.push('s');
+}
+```
+
+### BAD: Cloning when handling an `Option` or `Result`
+
+```rs
+pub struct LipogramCorpora {
+    selections: Vec<(char, Option<String>)>,
+}
+
+impl LipogramCorpora {
+    pub fn validate_all(&mut self) -> Result<(), char> {
+        for selection in &self.selections {
+            if selection.1.is_some() {
+                // BAD. prefer `if let Some(s) = selection.1.as_deref()` or pattern matching
+                if selection.1.clone().unwrap().contains(selection.0) {
+                    return Err(selection.0);
+                }
+            }
+        }
+        Ok(())
+    }
+}
+```
+
+### BAD: Cloning when passing immutable Struct into a closure
+
+If a `move` or `async move` closure needs immutable access to a Struct
+that is expensive to clone, it's good practice to share it via `Arc<_>`
+instead of `.clone()`.
+
+```rs
+// BAD
+let big_chunk = BigChunk::new();
+
+for _ in 1..10 {
+    let big_chunk = big_chunk.clone();
+    tokio::spawn(async move {
+        do_something_immutable(&big_chunk);
+    });
+}
+```
+
+```rs
+// GOOD
+let big_chunk = Arc::new(BigChunk::new());
+
+for _ in 1..10 {
+    let big_chunk = Arc::clone(&big_chunk);
+    tokio::spawn(async move {
+        do_something_immutable(&big_chunk);
+    });
+}
+```
+
+## Limitation
+
+Ignore redundant clones if their elimination would require adding explicit lifetime
+annotations (`'a`).
+
+<example>
+Using `.clone()` (Simple)
+The Config struct is "self-contained" and easy to pass around.
+
+```rs
+struct Config {
+    path: String,
+}
+fn load_config(path: &str) -> Config {
+    // Cloning ensures the Config owns its data
+    // and doesn't depend on the 'path' reference.
+    Config {
+        path: path.to_string(),
+    }
+}
+```
+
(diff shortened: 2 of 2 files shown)


TheSven73:
I believe the review-pr skill is mainly used by @derekbarbosa 

Below are a couple of recent commits/PRs where the skill could have provided value. Output from the updated /review-pr skill:

### Partitioning/Filtering Findings with `.clone()` (`7fba3a0` & `325c9f0`)
* **Impact**: **High**
* **Why it's a violation**: The codebase parses the `findings` array from AI outputs and keeps references `&[Value]`. When formatting the email body or printing CLI summaries, it clones each `Value` in `findings` to build a filtered list (such as splitting preexisting vs new bugs). 
* **The structural fix**: The finding values are large (they contain string descriptions and complete multi-line code snippets inside the `locations` key). Cloning them is expensive. Refactoring the temporary containers to hold references (`Vec<&Value>` instead of `Vec<Value>`) completely eliminates all heap allocations and copies:
```diff
-let mut new_findings = Vec::new();
-let mut preexisting_findings = Vec::new();
+let mut new_findings: Vec<&Value> = Vec::new();
+let mut preexisting_findings: Vec<&Value> = Vec::new();

 for f in findings_arr {
     if preexisting {
-        preexisting_findings.push(f.clone());
+        preexisting_findings.push(f);
     } else {
-        new_findings.push(f.clone());
+        new_findings.push(f);
     }
 }
```

### Cloning `entry.commit` inside the Polling Loop (`5d7e95d`)
* **Impact**: **Medium**
* **Why it's a violation**: Inside `src/bin/benchmark.rs`, the benchmark tool runs a polling loop that queries the database every 5 seconds for each commit. Inside the query parameters:
```rs
"SELECT id FROM patches WHERE message_id = ?",
libsql::params![entry.commit.clone()],
```
This is inside `for entry in &benchmark_entries`. The code clones `entry.commit` (a `String`) for the query parameters.
* **The structural fix**: SQLite/libsql's `params!` macro easily accepts borrowed references (e.g. `&entry.commit` or `entry.commit.as_str()`). Passing references avoids allocating new string buffers every 5 seconds:
```diff
-SELECT id FROM patches WHERE message_id = ?
-libsql::params![entry.commit.clone()]
+SELECT id FROM patches WHERE message_id = ?
+libsql::params![&entry.commit]
```

derekbarbosa:
@TheSven73

it is something that I have been meaning to fix somehow -- so thank you for addressing it through this PR and updating the agent skill.

This speaks to a different symptom though, @rgushchin, we should really be trying to enforce more robust lints via clippy and the buildsystem, which does a great job at picking up on these kind of anti-patterns.

Having stricter lints, cleaning up the "quality" (gross-over-generalization) of the code may improve `/pr-review` or however a user may "vet" the quality of their own changes via LLMs, etc. Something like this is caught by the perf and some of the redundant_* lints in clippy as-is.

I think we should look to how we can catch these using existing tools (clippy, llvm-cov +  [cargo-crap](https://github.com/minikin/cargo-crap)) that are cheap and easy to run in CI. Unless we want to spin up a sashiko-for-sashiko, teaching it to avoid antipatterns in Rust ;) 

I have opened a TODO item that may take quite some time to resolve here: #226

TheSven73:
@derekbarbosa good point, it definitely seems preferable to use deterministic linting tools such as clippy to find these classes of violations.

Is clippy’s `rendundant_clone` able to find most redundant clones, though?

I added this to Cargo.toml:
```toml
[lints.clippy]
redundant_clone = "warn"
```
<details>
<summary>clippy finds only two violations</summary>

```
    Checking sashiko v0.1.8 (/workspaces/melody/sashiko)
warning: redundant clone
   --> src/reviewer.rs:561:76
    |
561 |                     let prompts_hash_clone = prompts_hash.clone().map(|s| s.to_string());
    |                                                                            ^^^^^^^^^^^^ help: remove this
    |
note: this value is dropped without further use
   --> src/reviewer.rs:561:75
    |
561 |                     let prompts_hash_clone = prompts_hash.clone().map(|s| s.to_string());
    |                                                                           ^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
    = note: requested on the command line with `-W clippy::redundant-clone`

warning: `sashiko` (lib) generated 1 warning (run `cargo clippy --fix --lib -p sashiko` to apply 1 suggestion)
warning: redundant clone
   --> src/bin/benchmark.rs:600:41
    |
600 |             ("UNKNOWN".to_string(), text.clone())
    |                                         ^^^^^^^^ help: remove this
    |
note: this value is dropped without further use
   --> src/bin/benchmark.rs:600:37
    |
600 |             ("UNKNOWN".to_string(), text.clone())
    |                                     ^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
    = note: requested on the command line with `-W clippy::redundant-clone`

warning: `sashiko` (bin "benchmark") generated 1 warning (run `cargo clippy --fix --bin "benchmark"` to apply 1 suggestion)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 12.44s
```

</details>

<details>
<summary>
Yet Gemini 3.5 Flash found 24 violations (6% of all clones) using clone-discipline.md
</summary>

```md
   Category                                        │ Count │ Code Examples & Context
  ─────────────────────────────────────────────────┼───────┼────────────────────────────────────────────────────────────────────────────────
   1. Option/Result Handling Anti-Patterns         │  10   │ Cloning an  Option<String>  to call  .unwrap()  or  .unwrap_or()  instead of
                                                   │       │ borrowing with  .as_deref() . found in  src/reviewer.rs ,  src/bin/review.rs , 
                                                   │       │ src/bin/benchmark.rs ,  src/ai/bedrock.rs , and  src/ai/claude.rs .
   2. Mutually Exclusive Match / If-Else Arm Moves │   6   │ Redundant clones inside mutually exclusive  match  arms or  if / else 
                                                   │       │ branches (e.g.,  sashiko-cli.rs  and  db.rs ), where owned variables can be
                                                   │       │ safely moved/consumed directly in each branch.
   3. Redundant Owned-Value Clones (Last Use)      │   7   │ Cloning an owned variable when it is never used again in the remainder of the
                                                   │       │ function (e.g., a massive diff  Option<String>  clone in  src/db.rs , error
                                                   │       │ path variables in  src/api.rs , last-use query parameters in  src/inspector.rs
                                                   │       │ , and loop variables in  src/db.rs ).
   4. Unnecessary Loop cc Collection Clones        │   1   │ Cloning strings one-by-one in a loop to populate another collection in 
                                                   │       │ src/email_router.rs , when the source vector  p.cc  is already owned and can
                                                   │       │ be drained/extended directly ( cc.extend(p.cc) ).
   5. Redundant JSON Value Clones                  │   1   │ Cloning the  "payload"  field from an owned  serde_json::Value  (which could
                                                   │       │ contain a massive AI response body) in  src/ai/mod.rs  when it can be moved
                                                   │       │ via  .take() .
   TOTAL CONFIRMED VIOLATIONS                      │  24   │ 6.1% of the 389  .clone()  calls in the codebase.
```

</details>

Do LLM reviews provide more bang-for-the-buck? They might be more prone to false positives, although I have not yet seen Gemini 3.5 Flash flag false redundant clones. Perhaps some risk of false positives is acceptable, as long as it helps eliminate more redundant clones?

I suspect that unlike complexity checks, CRAP, etc, redundant clones are not optimized away by the Rust compiler, and may have real-life performance impact in terms of memory footprint, fragmentation, etc.

derekbarbosa:
Hi @TheSven73 -- you raise a lot of good points here. Also, thank you for the writeup + analysis!

> Is clippy’s rendundant_clone able to find most redundant clones, though?

I am by no means an expert on clippy's linting system, and I won't pretend to be a Rust expert here. In most of my day-to-day (and pre-LLM) usage, it realy has been a "set and forget" in my .toml files, along with some level of careful `.clone()` everywhere (things that LLM-gen'd code can ignore at times). 

IIUC the redundant_clone lint is actually pretty _conservative_, leading to false-negatives. I think this was done historically to prevent lint fixes from breaking working code. I am probably _very_ wrong here, but lints primarily interact with the HIR /AST data (type checking, type traits, validation) whereas borrow checking is handled primarily in MIR (borrow checking and lifetime validation). .clone() sits in between somewhere there, I guess -- It's somewhat confusing to me, as there are a good deal of steps between HIR and LLVM-IR.

Also, it looks like the additions to the skill are working, it looks great! :D 

> Do LLM reviews provide more bang-for-the-buck?

a pretty loaded question, but I will naively take a crack at it ;)

As I said before, I think the addition of the cloning discipline is a great one to the review skill. I think for performance and "safety" critical areas such as redundant_clone, we should probably strive to **continue** to add those, refining them and "fine-tuning" them; it can simplify the ease of code "checks" that can be replicated by something like a review bot with token access if/when we decide to set one up on the repository.

Generally speaking, LLMs do this kind of thing very well:  tokenizing massive amounts of text, performing some semantic checks with some "stored" prompt against it's trained parameters, and doing so (in the case of Flash, Sonnet, etc) very very fast. I am certainly not against using the right tool for the job, and in this case, we see that we stand to benefit a great deal from employing a tool that can do this reasonably well, and much faster than human eyes.

But I don't think it is a mutually exclusive effort -- striving to "better" the state of code. It should exist as a "quality gate" or a minimal standard of enforcement (at compile time) prior to review --similar to how a lot of projects do with humans ATM-- which is a "benefit" so-to-speak of using a language like Rust; We stand to benefit from a positive feedback loop if we employ both.

PR #228, opened 2026-05-29
fix: fix patch application

derekbarbosa:
Fix patch application failures and add diagnostics

Investigation of patchsets stuck in Failed To Apply status revealed four distinct root causes in the reviewer pipeline:

1. Patchsets with base-commit: trailers referencing subsystem tree SHAs that exist in the local object store but are not reachable from any named ref. git worktree add rejects these even though git rev-parse resolves them, causing the highest-priority baseline candidate to fail silently before any patch application is attempted.
2. Patches targeting trees that have diverged slightly from all available baselines. Both git am and git apply reject these due to strict context matching, even when the actual patch hunks are unambiguous.
3. 40 [GIT PULL] maintainer pull requests being ingested as patchsets. These contain git URLs and branches, not diffs, and always fail application. 29 were still pending review.
4. Version references like 7.1 and 7.2 extracted from subjects (e.g., "Fixes for 7.1", "sched_ext/for-7.2") producing baseline candidates for tags that don't exist yet during release candidate phases. Harmless but noisy in attempt logs.

To fix this, we employ the following:

- Worktree creation now retries at HEAD when the target SHA is unreachable from any named ref
- Added git apply -C2 as a third fallback after git am and git apply for context drift, tagged as git-apply-C2 in review metadata.
- [GIT PULL] messages are excluded from patchset classification at parse time
- Version tag baseline candidates (e.g., v7.1) are verified locally before being added, eliminating spurious resolution failures during RC phases

along with more logging on failure failure/success paths.

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -257,6 +257,11 @@ impl BaselineRegistry {
                     .is_ok()
                 {
                     candidates.push(BaselineResolution::LocalRef(tag_ref));
+                } else {
+                    info!(
+                        "Skipping version tag candidate {} (not found locally)",
+                        tag_ref
+                    );
                 }
             } else {
                 let tag_ref = format!("v{}", version);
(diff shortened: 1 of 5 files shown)


derekbarbosa:
@rgushchin gentle ping for review. let me know if this fix is no longer relevant so I can close.

PR #230, opened 2026-05-30
ai: restrict base URL paths with strict whitelisting

Rubusch:
Refactor `normalize_base_url` to enforce a strict whitelist of OpenAI endpoint paths using zero-dependency string parsing.

The previous logic blindly appended `/chat/completions` to any unrecognized suffix, creating malformed paths like `/v1/models/chat/completions`. The updated function isolates the path component and explicitly rejects anything other than empty, `/v1`, or `/v1/chat/completions` by returning an empty string. Extend test coverage accordingly.

Fixes: 9040f901d294 ("ai: fix OpenAI-compatible base URL")

### DIFF
--- a/src/ai/openai.rs
+++ b/src/ai/openai.rs
@@ -193,12 +193,23 @@ impl OpenAiCompatClient {
     /// path.  Our `post_request` POSTs directly to `self.base_url`, so we
     /// ensure the full path is present.
     fn normalize_base_url(url: &str) -> String {
-        let url = url.trim_end_matches('/');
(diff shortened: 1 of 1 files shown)


Rubusch:
Hi guys, please note, I don't want to be offensive implying with a "Fixes"...

The url fix for v1 and v1/chat/completions is working for my OLLAMA setup, thank you. Anyway, I just noticed, it could still be extended a bit more, catching bogus/wrong urls. When setting sashiko up locally it can be a bit more helpful to catch those things early.

PR #233, opened 2026-06-01
cli: surface the review subprocess's stderr when it fails

ftabba:
`sashiko-cli local` runs the review in a subprocess and streams its stderr only to drive the four-phase progress display, discarding every other line. When the subprocess failed before producing a result, the user saw only "Review subprocess produced no output (exit code: N)" with the actual cause (a stage error, an auth or quota failure, a panic) silently dropped.

Keep a bounded tail (the last 100 lines) of the subprocess's stderr while rendering progress, and replay it on the no-output failure path. Successful runs are unchanged (clean progress), while failures now explain themselves.

Assisted-by: Antigravity:gemini-3.1-pro
Signed-off-by: Fuad Tabba <tabba@google.com>

### DIFF
--- a/src/bin/sashiko-cli.rs
+++ b/src/bin/sashiko-cli.rs
@@ -1585,9 +1585,16 @@ async fn handle_local(
             drop(stdin);
         }
 
-        // Stream stderr for progress in a background task
+        // Stream stderr for progress, while keeping a bounded tail of the
+        // subprocess's log lines. The progress UI consumes these lines, so
+        // without the tail a failed review would collapse to a bare "no output"
+        // message with its actual cause (a stage error, an auth/quota failure)
+        // discarded. On failure we replay the tail so the run explains itself.
         let stderr = child.stderr.take();
         let stderr_handle = tokio::spawn(async move {
+            use std::collections::VecDeque;
+            const STDERR_TAIL_LINES: usize = 100;
+            let mut tail: VecDeque<String> = VecDeque::with_capacity(STDERR_TAIL_LINES);
             if let Some(stderr) = stderr {
                 use tokio::io::{AsyncBufReadExt, BufReader};
                 let reader = BufReader::new(stderr);
@@ -1609,8 +1616,13 @@ async fn handle_local(
                         eprint_phase(4, 4, "Review complete.");
                         eprintln!();
                     }
+                    if tail.len() == STDERR_TAIL_LINES {
+                        tail.pop_front();
+                    }
+                    tail.push_back(line);
                 }
             }
+            tail
         });
 
         // Capture stdout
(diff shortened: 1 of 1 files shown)


derekbarbosa:
output of review-pr:


🟢 Design Alignment
Faithful to the local command intent in designs/DESIGN_CLI_TOOL_V2.md (clear, self-explanatory UX). The change targets exactly the failure mode described: subprocess aborts before emitting a result (review.rs:82-85 panic hook + default panic output → empty stdout, exit 101). No design conflict. The bounded-tail approach (last 100 lines) correctly captures the causal error since panics/backtraces land at the end of stderr.

🟡 Complexity / Logic — replay only fires on empty stdout
src/bin/sashiko-cli.rs:1639 gates the replay on stdout.trim().is_empty(). The subprocess also fails softly with non-empty stdout, in which case the stderr tail is still silently dropped:
- Error JSON printed to stdout with exit 0 (review.rs:431, review.rs:451) — e.g. "Patch application failed", AI provider failures. The error string is surfaced, but the detailed stderr logs (retries, provider exceptions) remain swallowed.
- Partial/corrupt stdout → the parse at line 1653 fails with "Failed to parse review output JSON" and the tail is discarded right where it would help most.
```
Fix — replay the tail on the parse-failure path too:
-        // Parse the review output
-        let result: Value =
-            serde_json::from_str(stdout.trim()).context("Failed to parse review output JSON")?;
+        // Parse the review output
+        let result: Value = match serde_json::from_str(stdout.trim()) {
+            Ok(value) => value,
+            Err(err) => {
+                for line in &stderr_tail {
+                    eprintln!("{line}");
+                }
+                return Err(anyhow::anyhow!(
+                    "Failed to parse review output JSON: {} (exit code: {})",
+                    err,
+                    exit_code
+                ));
+            }
+        };
```

🔵 Style / Nits / DRY — replay loop + single write
src/bin/sashiko-cli.rs:1642-1644 does one eprintln! syscall per line. A single write is idiomatic for a bounded 100-line buffer:
```
         if stdout.trim().is_empty() {
             // The subprocess failed before emitting a result; replay its log
             // tail so the actual cause is visible rather than swallowed.
-            for line in &stderr_tail {
-                eprintln!("{line}");
+            if !stderr_tail.is_empty() {
+                eprintln!(
+                    "{}",
+                    stderr_tail.iter().map(|s| s.as_str()).collect::<Vec<_>>().join("\n")
+                );
             }
             return Err(anyhow::anyhow!(
```

🔵 Test Coverage — no test for the new path
handle_local's subprocess logic has zero unit/integration coverage (no tests under tests/ touch it). The new tail/replay behavior is exactly the kind of regression-prone plumbing worth pinning down with an integration test that points the CLI at a stub sashiko-review binary that writes N log lines to stderr and exits non-zero with empty stdout, asserting the tail is replayed. Given GEMINI.md's testing requirement for new logic, this is the main gap.


Notes (non-actionable)
- stderr_handle.await.unwrap_or_default() (line 1634) silently degrades to the old behavior if the drain task panics — acceptable, but a comment would clarify intent.
- Tail is bounded to 100 lines; a verbose failure that logs the cause early and then churns ≥100 lines could push the real error out of the window. Edge case only; the panic-backtrace case (the target) is unaffected.
- No introduced regressions found: no deadlock risk (stderr pipe is drained concurrently on the same runtime; wait_with_output only touches stdout since stderr was take()n), no success-path behavior change, no clone/DRY issues, exit-code semantics unchanged.

ftabba:
Superseded by #514, which carries this commit unchanged, adds the stage progress lines the tail was plumbing for, and replays the tail on the parse-failure path the review comment above raised.

PR #235, opened 2026-06-02
RFC: Add bounded local model review

anders-heimer:
This RFC adds a bounded review mode for OpenAI-compatible local models and
tightens the review/benchmark accounting needed to evaluate that mode.

The main goal is to keep the provider selection as a transport choice while
allowing local models to opt into Sashiko-controlled review behavior: bounded
tool exploration, schema-only finalization, token preflight checks, typed budget
failures, and minimal fallback output when the full staged review cannot fit.

The series also preserves Stage 8/Stage 9 accounting through finalization and
benchmark reporting, so retained concerns, dropped candidates, final findings,
and failed/skipped rows are easier to audit.

I am currently running the 1000-patch benchmark on an L40S:

Reviews:
Failed    274
Reviewed  545
Skipped     7
In Review   1

Patch statuses:
Reviewed      545
FailedBudget  271
Failed          3
Skipped         7
Pending/null   60

Benchmark configuration:

benchmark=benchmarks/benchmark.json
repo=third_party/linux
provider=openai-compatible
model=qwen3-coder-30b-a3b-128k-t0
bounded_local_model=true
max_input_tokens=90000
context_window_size=131072
max_tokens=6144
temperature=0.0
api_timeout_secs=1200
review_concurrency=1
review_timeout_seconds=7200
review_max_retries=0
enable_static_bug_seeds=true
enable_targeted_bug_pattern_prescan=false


rgushchin:
Do you get any useful results out of it? Any benchmark results I can share?
I've mixed feelings about supporting various limited modes: from one perspective I can see how they are useful, but at the same time they increase the testing surface and their long-term value is kinda unclear. To me the decision factor is how actually useful they actually are.

anders-heimer:
Hi Roman,

Yes, I get some useful results, but it is far from frontier-level quality.

The way I think about it is somewhat similar to risc-v, we do not care about it because every risc-v board beats x86/arm systems today.

On the 999-entry benchmark, the setup produced 33 exact detections and 143 strong partials: 176/999 actionable useful signal end-to-end, or 176/640 among valid reviewed reports, using Opus as the judge. The main limitation is still budget/model quality, with 339 budget failures and 7 protocol failures.

Attached the result of the benchmark.

Thanks, Anders
[review-comments.html](https://github.com/user-attachments/files/28802537/review-comments.html)

derekbarbosa:
Hi, is there any consensus on this approach @rgushchin @anders-heimer ?

anders-heimer:
Hi,

I will run the benchmarks again to compare both how the latest "open source" models behave and how the current Sashiko code base behaves. I expect that I will see that improvements have been made in both. 

Br, Anders

PR #273, opened 2026-06-12
docker: Build with persistent cache

Rongronggg9:
Cache downloaded and generated files with `RUN --mount=type=cache`, so that later builds don't need to download and build everything from scratch.

Some Docker versions could run into bugs when a later stage accesses a cached directory from previous stages, so copy the binaries into a separate directory first. Doing so also allows Stage 2 to merge multiple `COPY --from=builder` steps into a single one.

This series also bumps rust image version to 1.90, otherwise it keeps downloading 1.90 according to `rust-toolchain.toml`.

Link: https://docs.docker.com/build/cache/optimize/#use-cache-mounts

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,17 @@ COPY . .
 ## This makes the image large but ensures fast startup in Cloud Run
 #RUN wget -c https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/clone.bundle -O /usr/src/sashiko/linux-kernel.bundle
 
-# Build for release
-RUN cargo build --release
+# Build for release with persistent cache
+# Note: some Docker versions could run into bugs when a later stage accesses a
+# cached directory from previous stages, so copy the binaries into a separate
+# directory first.
+RUN --mount=type=cache,target=/usr/src/sashiko/target/ \
+    --mount=type=cache,target=/usr/local/cargo/git/db/ \
+    --mount=type=cache,target=/usr/local/cargo/registry/ \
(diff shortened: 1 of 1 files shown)


Rongronggg9:
Added a new commit to cache the installations of `clippy` and `rustfmt` as well.

rgushchin:
Thank you for the PR! What's the main benefit of enabling the cache? Tbh, I don't really see the point: I build the docker image few times a week and it's reasonably fast, so... But maybe I'm missing something?

Rongronggg9:
Thanks for reviewing this!

Some regions in the world have poor connections to Docker Hub and crates.io. Redownloading everything takes more than ten minutes and is just painful.

Of course, it's possible to modify the Dockerfile to enable mirrors (I did!). However, the download and compilation time still adds up quickly and becomes frustrating when I need to revise and test repeatedly. I am preparing a series to improve support for the OpenAI completions API, and sometimes unit tests can't cover obscure divergences between different API providers, so I have to test with real APIs[^1]. The frustration during the process led me to try enabling caching. Now I only need to wait for less than one minute in the revise-and-test process, and I am happy with that.

[^1]: Why do I need to test it in a container? Well, I've already deployed Sashiko on my home server, so that's the best place for test in production (TM). And I don't want to let Sashiko mess up with my working Linux repository on my PC or create another clone for testing Sashiko.

derekbarbosa:
@rgushchin would this still be worth pulling in?
it appears to be harmless to those unaffected, and a net win for the previously described conditions.

PR #277, opened 2026-06-15
feat: build parent message chain in References header with cycle protection

davidehu-69:
Outgoing Sashiko review replies were only containing the immediate parent Message-ID in the `References` and `In-Reply-To` headers. For deep email threads, this broke thread threading in several mail user agents (MUAs), causing review messages to appear outside the main thread or fragmented.
### Solution
This PR implements full parent message chain retrieval to populate the `References` header, restoring RFC 5322 compliant threading while adding safety limits to prevent database traversal loops.
* **Database Lookup (`src/db.rs`):** 
  * Implemented `get_message_references_chain` to recursively walk up the parent `in_reply_to` chain.
  * Added cycle detection (via a `visited` set) and a max depth cutoff of `100` to prevent infinite loops from malformed or cyclic mail headers.
* **Review Handler (`src/reviewer.rs`):**
  * Retrieved the parent references chain and formatted it as a space-separated string of bracketless Message-IDs, passing the immediate parent as `in_reply_to`.
* **Mail Worker (`src/worker/email.rs`):**
  * Split the outbox references string, wrapped each Message-ID in angle brackets (`<>`), and passed the sequence to the Lettre email builder.
* **Tests (`src/db.rs`):**
  * Added regression and unit tests covering standard chain building, multi-message cyclic loop prevention, and self-referencing loop prevention.
### References
* **Mailing List Thread:** [Incorrect/lacking References in mail headers?](https://lore.kernel.org/sashiko/20260519-google-salary-aab923033a86@spud/) on `lore.kernel.org`

### Alternative Designs & Future Optimizations (Not Implemented)
To optimize performance, we evaluated two alternative architectures but decided against them for the following reasons:
1. **Recursive SQL Common Table Expressions (CTEs):**
   * *Concept:* Retrieve the entire chain in a single database roundtrip using a recursive SQLite query.
   * *Trade-off:* Offloading cycle detection to a recursive CTE increases SQL complexity significantly. More importantly, recursive CTE syntax and cycle features can vary or fail on older/varying SQLite/libsql versions in the deployment environment. We opted for explicit Rust traversal to maximize portability and readability.
2. **Pre-computed Materialized Paths (Store References at Ingestion):**
   * *Concept:* Calculate and store the full references chain directly in a column on the `messages` table at ingestion time, reducing read-time lookup to $O(1)$.
   * *Trade-off:* If emails are ingested out-of-order (common in distributed mailing list fetching), backfilling the path links would require complex write-time stitching logic. Given that thread depths should be typically small ($<20$ messages) and SQLite queries are extremely fast, the runtime overhead of traversing $N$ indexed queries in Rust is negligible compared to the added write-time complexity.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -86,6 +86,7 @@ pub struct MessageRow {
     pub git_blob_hash: Option<String>,
     pub mailing_list: Option<String>,
     pub diff: Option<String>,
+    pub references_hdr: Option<String>,
 }
 
 pub struct AiInteractionParams<'a> {
@@ -175,7 +176,7 @@ impl Database {
 
     pub async fn get_message_details(&self, id: i64) -> Result<Option<MessageRow>> {
         let mut rows = self.conn.query(
-            "SELECT m.id, m.message_id, m.thread_id, m.in_reply_to, m.author, m.subject, m.date, m.body, m.to_recipients, m.cc_recipients, m.git_blob_hash, m.mailing_list, p.diff 
+            "SELECT m.id, m.message_id, m.thread_id, m.in_reply_to, m.author, m.subject, m.date, m.body, m.to_recipients, m.cc_recipients, m.git_blob_hash, m.mailing_list, p.diff, m.references_hdr 
              FROM messages m 
              LEFT JOIN patches p ON m.message_id = p.message_id
              WHERE m.id = ?",
@@ -197,6 +198,7 @@ impl Database {
                 row.get::<Option<String>>(10).ok().flatten(),
                 row.get::<Option<String>>(11).ok().flatten(),
                 row.get::<Option<String>>(12).ok().flatten(),
+                row.get::<Option<String>>(13).ok().flatten(),
             ))
         } else {
             None
@@ -216,6 +218,7 @@ impl Database {
             git_blob_hash,
             mailing_list,
             raw_diff,
+            references_hdr,
         )) = row_data
         {
             // Fetch thread messages
@@ -260,6 +263,7 @@ impl Database {
                 git_blob_hash,
                 mailing_list,
                 diff,
+                references_hdr,
                 thread: Some(messages),
             }))
         } else {
@@ -427,6 +431,9 @@ impl Database {
             .await;
         let _ = self
             .try_add_column("messages", "mailing_list", "TEXT")
+            .await;
+        let _ = self
+            .try_add_column("messages", "references_hdr", "TEXT")
             .await;
         let _ = self
             .try_create_index(
@@ -1572,6 +1579,39 @@ impl Database {
         git_blob_hash: Option<&str>,
         mailing_list: Option<&str>,
     ) -> Result<()> {
+        self.create_message_with_references(
+            message_id,
+            thread_id,
+            in_reply_to,
+            author,
+            subject,
+            date,
+            body,
+            to,
+            cc,
+            git_blob_hash,
+            mailing_list,
+            None,
+        )
+        .await
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub async fn create_message_with_references(
+        &self,
+        message_id: &str,
+        thread_id: i64,
+        in_reply_to: Option<&str>,
+        author: &str,
+        subject: &str,
+        date: i64,
+        body: &str,
+        to: &str,
+        cc: &str,
+        git_blob_hash: Option<&str>,
+        mailing_list: Option<&str>,
+        references_hdr: Option<&str>,
+    ) -> Result<()> {
         // Check for thread merge (Thread split resolution)
         if let Ok(Some(old_thread_id)) = self.get_thread_id_for_message(message_id).await
             && old_thread_id != thread_id
@@ -1629,8 +1669,8 @@ impl Database {
         // But main.rs logic should ensure consistency.
         // Use INSERT OR REPLACE.
         self.conn.execute(
-            "INSERT INTO messages (message_id, thread_id, in_reply_to, author, subject, date, body, to_recipients, cc_recipients, git_blob_hash, mailing_list) 
-             VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+            "INSERT INTO messages (message_id, thread_id, in_reply_to, author, subject, date, body, to_recipients, cc_recipients, git_blob_hash, mailing_list, references_hdr) 
+             VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
              ON CONFLICT(message_id) DO UPDATE SET
                 thread_id=excluded.thread_id,
                 in_reply_to=excluded.in_reply_to,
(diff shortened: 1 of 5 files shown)


TheSven73:
Davide, thanks a lot for your contribution !

Have you considered simpler solutions? For example - when a email client replies, it takes the References: header from the parent email, copies it exactly, and appends the parent's Message-ID to the very end of the list. That ensures RFC 5322 compliance.

In sashiko this could mean:
- extend the database schema to store `references` in the message table, and
- when inserting into the email_outbox table, set `references` to `parent_message_references + msg_id` (instead of simply `msg_id` as happens today)

Perhaps it would be useful to compare tradeoffs between solutions?

davidehu-69:
Hi Sven,

Thanks for the feedback! I was initially hesitant to introduce a database schema change, but your suggestion is indeed much cleaner, faster, and more robust.

I have updated the PR to adopt this approach:
- **Schema Migration:** Added a `references_hdr` column to the `messages` table. Sashiko should handle the migration on startup.
- **O(1) Resolution:** Replies now read the parent's `references_hdr` directly and append the parent's `Message-ID`. This runs in $O (1)$ query time and completely eliminates the recursive traversal/cycle-protection code, making the logic much simpler and cleaner.
- **Safety & Fallback:** Added unit tests covering the `NULL`/`None` record path to ensure the migration is safe for historical messages (which fallback to parent-only references, matching the old behavior).

Let me know what you think.

TheSven73:
Thanks ! I believe the new, simplified commit has issues with handling of angle brackets, but I suggest we wait until @rgushchin has had a chance to weigh in.

rgushchin:
Thank you! I like the new simpler design better. The commit message lacks the description and the SOB though and also the linter is not entirely happy. Please, make sure to include the description on how the change was tested (including the db schema migration). Thanks!

src/reviewer.rs
```diff
@@ -2098,6 +2096,7 @@ impl Reviewer {
         let findings_count = findings.map(|f| f.len()).unwrap_or(0);
 
         let msg_id = patch_message_id;
+        let msg_id_clean = msg_id.trim_matches(|c| c == '<' || c == '>');
```
TheSven73, line 2101: I thought this commit had angle bracket issues, because I got confused by `msg_id` / `msg_id_clean` .

`msg_id` is passed directly to `get_message_details_by_msgid` which leads to a db query, suggesting that the id is already clean. But then further down the function, it's explicitly cleaned.

This is not a problem of this PR, so let's disregard.

src/worker/email.rs
```diff
@@ -105,10 +105,13 @@ impl EmailWorker {
...
-                "<{}>",
-                email_row.references_hdr
-            )));
+            let refs: Vec<String> = email_row
+                .references_hdr
+                .split_whitespace()
+                .map(|part| format!("<{}>", part))
+                .collect();
+            let refs_str = refs.join(" ");
+            builder = builder.header(lettre::message::header::References::from(refs_str));
```
TheSven73: Nit: could be
```rs
builder = builder.references(refs_str);
```
?

src/schema.sql
```diff
@@ -45,6 +45,7 @@ CREATE TABLE IF NOT EXISTS messages (
     cc_recipients TEXT,
     git_blob_hash TEXT,
     mailing_list TEXT,
+    references_hdr TEXT,
```
TheSven73, line 48: Nit: why `references_hdr` and not `references` ?

src/reviewer.rs
davidehu-69: Thanks for checking! Yes, the raw vs. clean Message-ID handling is a bit fragmented across the code. I will create an issue to track this and fix it in later PR. What do you think?

src/worker/email.rs
davidehu-69: I checked the `lettre` docs: `MessageBuilder` has built-in helpers for `to`, `from`, and `subject`, but it doesn't expose a `.references()` method directly. It has to be set as a typed header via `.header(lettre::message::header::References::new(...))` or `.header(...)` as done here, otherwise it won't compile.

src/schema.sql
davidehu-69: Unfortunately, `REFERENCES` (case-insensitive) is a reserved keyword in SQL (used for foreign key constraints). To avoid syntax conflicts in the database queries, I had to find a different name; `references_hdr` (or alternatively `refs`) seemed like the clearest way to represent it. Let me know if you had a better name in mind.

src/worker/email.rs
TheSven73: Using `references()` compiles for me, and is also present in the lettre [docs](https://docs.rs/lettre/latest/lettre/message/struct.MessageBuilder.html#method.references) . Not sure what is different for you?

davidehu-69:
Hi Sven and Roman,

Thanks again for checking this PR and helping me out with my first contribution on Sashiko.

I have updated the PR branch:
- Squashed the cargo formatting changes directly into the main commit.
- Addressed the linter warnings in `src/worker/tools.rs` by adding `#[allow(clippy::collapsible_match)]` to the helper function in a separate, second commit. This silences the warnings while preserving consistent block statement style across all match arms. This linter warning does not come from my changes, so I kept it as a separate commit.
- Concerning testing, migration happens on startup and I introduced a test for backward compatibility.
- Added commit description and SOB lines to both commits. 
- Ran the PR checks locally via `make check-pr`, and all tests, formatting, clippy, and DCO checks are now passing.

src/schema.sql
TheSven73: You could use `"references"` which escapes the reserved keyword issue. But this is only a nit, it has little importance, follow your own preference.

src/reviewer.rs
TheSven73: I think it's very minor, and entirely up to @rgushchin to decide if it's even worth creating an issue for. I suspect there are much more urgent and important things to address.

src/worker/email.rs
davidehu-69: My apologies. You are right. I think I checked something else. I will fix it.
Thanks for also pointing directly to correct docs.

src/worker/tools.rs
```diff
@@ -214,6 +214,7 @@ impl ToolBox {
         decls
     }
 
+    #[allow(clippy::collapsible_match)]
```
TheSven73: Not needed on my end - my clippy does not generate any warnings here. Why is it different for you?

```
$ cargo clippy --version
clippy 0.1.90 (1159e78c47 2025-09-14)
```

src/worker/email.rs
davidehu-69: Fixed.

src/db.rs
```diff
@@ -5914,4 +5956,85 @@ mod tests {
         let length: i64 = row.get(0).unwrap();
         assert_eq!(length, 456);
     }
+
+    #[tokio::test]
+    async fn test_message_references_header_storage() {
```
TheSven73, line 5961: If I understand correctly, this verifies that references are handled correctly by the db.

But do you have test coverage to verify that generated review emails actually have proper references? Would it make sense to extend `test_queue_notifications_split_summary` to verify that `email_outbox` rows have the references you expect?

src/worker/tools.rs
davidehu-69:
```
$ cargo clippy --version
clippy 0.1.95 (59807616e1 2026-04-14)
```

I just saw in `rust-toolchain.tml` that Sashiko is using:

```
[toolchain]
channel = "1.90.0"
components = ["rustfmt", "clippy"]
```

Again, sorry for the mistakes and I am super grateful for your help. I will fix it and drop the unnecessary commit.

src/db.rs
davidehu-69: Hi Sven, you are correct.

Rather than extending `test_queue_notifications_split_summary`, I think it might be better to create a new dedicated unit test. The referenced test is focused on body formatting of split email summaries, I feel like coupling it with references header logic felt a bit crowded.

I added a new dedicated integration unit `test test_queue_notifications_email_references` in `src/reviewer.rs` that explicitly checks that the database outbox rows contain the correct `in_reply_to` and `references_hdr` sequence for both root messages (backward-compatibility path) and deep references chains.

PR #279, opened 2026-06-16
fix: reap timed-out review children

### LINKED ISSUE #248: Review children deadlock and timeout failure with worktree reuse
# Bug Report: Review children deadlock and timeout failure with worktree reuse

## Summary

After commit `bf158e7d` ("feat: optimize parallel reviews to reuse a single
worktree"), concurrent review children can deadlock permanently, and the
parent sashiko process fails to recover because:

1. The review children enter an unrecoverable state (all threads blocked on
   futex waits, zero syscall activity, no network connections).
2. The parent's 3600s timeout fires, but `child.wait().await` at
   `src/reviewer.rs:2062` has no timeout — it blocks forever on a
   deadlocked child.

Additionally, a secondary cache poisoning issue was discovered: empty AI
responses (0 output tokens) are cached and replayed on every retry,
turning transient failures into permanent ones.

## Observed behavior

Patchset 89 (5 patches) was submitted for review.  Sashiko spawned 3
concurrent `review` child processes for patches 1, 3, and 4, all sharing
a single worktree via `--reuse-worktree`.

### Deadlocked children (observed via /proc)

| PID    | Patch | Elapsed | State |
|--------|-------|---------|-------|
| 502592 | 1     | 1h20m+  | 34 threads, 32 on futex_do_wait, 1 on ep_poll, 1 on anon_pipe_read. Zero TCP connections. Zero syscalls over 5s strace. |
| 502626 | 4     | 1h20m+  | Identical to above. Also: `--review-commit` = `--baseline` (same SHA), meaning zero diff. |
| 505965 | 3     | 1h20m+  | Same, plus **4 zombie git child processes** (Z+ state, `<defunct>`). |

### Parent (PID 502088) also hung

- 33 threads, 32 on futex_do_wait, 1 on ep_poll.
- epoll_wait returning normally (timer wheel alive), but the tokio tasks
  reading from children's stdout were never waking up.
- All stdin/stdout/stderr pipes to children still open (verified via
  `/proc/502088/fd`).
- The 3600s timeout had clearly passed (children running 1h20m), but the
  parent never killed or reaped the children.

### Timeline

- Multiple review attempts (2937–2944) were created for patchset 89.
- Reviews 2940 and 2941 were left orphaned as "In Review" from the
  deadlock — never cleaned up by the restart recovery.
- Required a full machine reboot to break the deadlock.
- After reboot, the same patches hit a different failure: Gemini returning
  0 output tokens for Stage 11 (LKML report generation).  These empty
  responses were cached and replayed on every retry attempt, preventing
  any progress.

### 1. Race condition in worktree reuse (primary)

Commit `bf158e7d` passes the same worktree path to all concurrent review
children:

```rust
// src/reviewer.rs:622
Some(&worktree_path_clone), // Reuse the single worktree!
```

The commit message states it relies on "virtualized HEAD" to avoid
physical checkout.  However, when `commit_sha` is `None` (line 561),
the review binary falls back to physical worktree mutation:

```rust
// src/reviewer.rs:561
let commit_sha = patch_commits.get(index).cloned();
```

`patch_commits` can silently lose entries when `get_commit_hash()` fails
(line 950–952) — there is no error propagation, just a missing insert.
When this happens, `commit_sha` is `None`, and the review binary falls
back to `git am` / `git reset --hard` on the shared worktree (lines
468–530 in `src/bin/review.rs`), corrupting the working directory for
all concurrent children.

Even when `commit_sha` is present, the review binary still runs git
commands (`git_show`, `git_grep`, `git_log`) that read from the worktree.
While these are read-only, concurrent writes from a fallback child
corrupt their view.

The 4 zombie git processes under PID 505965 are direct evidence: the
review binary spawned git commands that completed, but the async code
waiting for their output deadlocked — likely because the worktree state
was mutated by a concurrent child between the spawn and the await.

### 2. Parent timeout does not recover stuck children (secondary)

After the interaction loop exits (timeout or completion), the parent
calls:

```rust
// src/reviewer.rs:2062
let _ = child.wait().await; // Reap zombie
```

This has **no timeout**.  If the child process is deadlocked and never
exits, `child.wait()` blocks forever, which in turn blocks the parent's
tokio task indefinitely.  The 3600s timeout on the interaction loop is
rendered useless.

**Fix**: call `child.kill()` before `child.wait()` when the interaction
loop exits via timeout, or wrap `child.wait()` in a
`tokio::time::timeout()`.

### 3. Empty responses cached permanently (tertiary)

When the AI provider returns a response with no content and no tool calls
(observed as `out=0` from Gemini), the cache stores it unconditionally:

```rust
// src/ai/cache.rs:397-427
let resp = self.inner.generate_content(request.clone()).await?;
// ... no validation ...
let _ = self.conn.execute("INSERT OR REPLACE INTO response_cache ...", ...).await;
```

On retry, the same cache key produces the same empty response, causing
the stage to fail identically on every attempt.  This was observed in
the logs:

```
Cache hit [448edfda5c77] — 9.125 tokens saved   ← poisoned entry
Gemini response received. Tokens: in=10386, cached=0, out=0
Stage 11 failed (attempt 1/3): No content or tool calls from AI
Cache hit [448edfda5c77] — 9.125 tokens saved   ← same poisoned entry
Stage 11 failed (attempt 2/3): No content or tool calls from AI
Cache hit [448edfda5c77] — 9.125 tokens saved   ← same poisoned entry
Stage 11 failed (attempt 3/3): No content or tool calls from AI
```

**Fix**: committed in `cache_lru_eviction` branch as `f05f558a`
("fix(cache): never cache empty AI responses").  Guards on both write
side (skip INSERT) and read side (DELETE + fall through to live provider).

### 4. Patch 4 baseline == review-commit anomaly

Patch 4 was launched with:
```
--baseline 0b8ed6305a54c7676291d06092d9991088da148b
--review-commit 0b8ed6305a54c7676291d06092d9991088da148b
```

Same SHA for both baseline and review-commit means zero diff.  This may
have contributed to the deadlock (review binary in an unexpected state)
and to Gemini returning empty Stage 11 responses (nothing to report).

## Likely culprit

**`bf158e7df9b785b828dc55c5ca973e4b4df15330`** — "feat: optimize parallel
reviews to reuse a single worktree" (Roman Gushchin, 2026-06-02).

The commit removes all physical checkout/reset operations from the review
binary and passes the shared worktree to concurrent workers, but does not
add synchronization and does not handle the fallback case where
`commit_sha` is `None`.

## Suggested fixes

1. **Add locking or per-child worktrees**: Either serialize worktree
   access with a mutex, or give each concurrent child its own lightweight
   worktree (git worktrees are cheap with `--detach`).

2. **Fail loudly when `commit_sha` is None**: Instead of silently passing
   `None` and falling back to physical checkout, emit an error and skip
   the patch.

3. **Kill children on timeout**: Replace the bare `child.wait().await`
   with `child.kill()` + `child.wait()` when the interaction loop exits
   via timeout.

4. **Propagate `get_commit_hash` errors**: Line 950 silently drops
   errors.  Make it log a warning and mark the patch as failed instead
   of leaving a gap in `patch_commits`.

## Environment

- Sashiko built from `ai_pricing` branch (stacked on `cache_lru_eviction`
  on `response_cache_stats` on `origin/main`).
- AI provider: `stdio-gemini`
- Platform: Fedora 44, Linux 7.0.10, running inside toolbx container.
- Rust 1.95.0

nkgotcode:
Fix review tool timeout handling so a child that stops producing output is terminated and reaped instead of leaving the parent waiting on process exit.

The timeout path now starts child termination and bounds the reap wait, and the regression test covers a child that ignores termination long enough to exercise the bounded wait.

Fixes #248

Validation:

- `cargo test test_run_review_tool_timeout_reaps_stuck_child -- --nocapture --test-threads=1`
- `cargo test test_run_review_tool_concurrency -- --nocapture --test-threads=1`

### DIFF
--- a/src/reviewer.rs
+++ b/src/reviewer.rs
@@ -1611,7 +1611,7 @@ async fn run_review_tool(
     use std::sync::atomic::{AtomicBool, AtomicU32, AtomicUsize, Ordering};
     use std::time::Duration;
     use tokio::time::Instant as TokioInstant;
-    use tokio::time::timeout_at;
+    use tokio::time::{timeout, timeout_at};
 
     // Perform interaction with timeout
     let deadline = Arc::new(std::sync::Mutex::new(
@@ -1991,7 +1991,23 @@ async fn run_review_tool(
     }
     drop(stdin_writer);
 
-    let _ = child.wait().await; // Reap zombie
+    let timed_out = match &interaction_result {
+        Err(e) => e
+            .to_string()
+            .contains("Review tool timed out (active time exceeded)"),
+        Ok(_) => false,
+    };
+
+    if timed_out {
+        error!(
+            "Review tool timed out after {} active seconds. Killing process.",
+            settings.review.timeout_seconds
+        );
+        let _ = child.start_kill();
+        let _ = timeout(Duration::from_secs(5), child.wait()).await;
+    } else {
+        let _ = child.wait().await; // Reap zombie
+    }
 
     match interaction_result {
         Ok(json) => {
@@ -2040,19 +2056,7 @@ async fn run_review_tool(
             }
             Ok(json)
         }
-        Err(e) => {
-            // Check if it's the specific active time exceeded error we throw in the loop
-            if e.to_string()
-                .contains("Review tool timed out (active time exceeded)")
-            {
-                error!(
-                    "Review tool timed out after {} active seconds. Killing process.",
-                    settings.review.timeout_seconds
-                );
-                let _ = child.kill().await;
-            }
-            Err(e)
-        }
+        Err(e) => Err(e),
     }
 }
 impl Reviewer {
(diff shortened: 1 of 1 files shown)


rgushchin:
Please, add your SOB line to the commit log.

nkgotcode:
Added the Signed-off-by trailer in ed15d07.

rgushchin:
Can you also please add a proper commit description and your real email address?
Basically I expect commit messages to follow the Linux kernel style.

Thanks!

nkgotcode:
Amended the commit in 7569a66 with a Linux-kernel-style description and real email metadata. The author, committer, and Signed-off-by trailer now use `Khanh Le <nkgotmail@gmail.com>`.

PR #293, opened 2026-06-25
prompts: add more details for the PCI subsystem

kwilczynski:
The PCI subsystem has some specific needs and wants, so capture some of these for Sashiko. Perhaps with the extra content, Sashiko will be able to capture PCI-specific issues as part of the review process.

### DIFF
--- a/third_party/prompts/kernel/subsystem/pci.md
+++ b/third_party/prompts/kernel/subsystem/pci.md
@@ -1,77 +1,500 @@
 # PCI Subsystem Details
 
+Everything below describes properties a well-formed PCI patch should
+satisfy.  When reviewing, CHECK the patch under review against these
+properties and FLAG deviations.  Do not edit, rewrite, or perform any of
+these actions yourself.  An imperative ("Add ...", "Strip ...", "Wrap ...")
+describes what the author should have done; flag it if they did not.
+Flag a violation based on lines the patch ADDS; you may read
+surrounding context to confirm whether a matching release or guard is
+present, but never flag pre-existing code for its own sake.
+
+Every finding has one of three dispositions:
+
+- **REPORT as bugs**: code-correctness invariants; report as a bug.
+- **Low-severity suggestion**: prose, spelling, formatting, style.
+- **Silent**: unsure, or evidence insufficient to confirm.
+
+When unsure whether a rule applies, or when the evidence needed to
+confirm a violation is not visible in the diff, stay silent.  Check
+the code-correctness
+invariants (the sections marked **REPORT as bugs**) before prose,
+spelling, and formatting items, so that under an output budget the
+bugs are reported first.  This file is a subsystem module loaded after
+`review-core.md`, which owns output format and Fixes-tag policy.
+The prose, spelling, and style sections below apply only when
+subjective review is enabled; render their findings per
+`inline-template.md`.  Every REPORT-as-bugs finding contributes at
+least `low` severity to the `review-core.md` Task 5 aggregate.
+These dispositions (REPORT as bugs, Low-severity suggestion, Silent)
+are internal triage labels; never emit them in the review output.
+Report each distinct defect once, even if it matches multiple
+sections.
+
+## Commit Message Title Format
+
+Using the wrong subject prefix or capitalization style makes the patch
+look out of place in `git log --oneline` output and forces the maintainer
+to silently rewrite it before applying, delaying the patch.
+
+Check the subject prefix against the conventions below and the path of the
+file being changed.  You have only the patch; do not claim to have run
+`git log` or compared against history.  The PCI subsystem uses
+hierarchical prefixes separated by colons:
+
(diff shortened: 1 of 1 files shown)


rgushchin:
Thanks, Krzysztof! Overall I like the change and it might make reviews better. Few concerns though:
1) Who is supposed to do what? (Target Ambiguity)
  The prompt uses commands like "Add a changelog" or "Strip timestamps". Because the AI is reading this, it might get
  confused and think it needs to do those things, rather than checking if the patch author did them. We should change
  these to "Verify that the author..." so the AI knows its job is just to check.
2) Checking things the sashiko can't see (Unverifiable Rules)
  Some rules ask the AI to check if the author waited a few days before reposting, addressed all previous feedback, or
  updated Patchwork. Sashiko only sees the single patch file and has no access to email history as of now. If we tell
  the AI to check these, it will just guess (hallucinate) or complain about things it can't actually verify. We should
  remove these workflow rules for now.
3) Some redundancy in quick checks (maybe it's ok, but worth thinking of)
  The "Quick Checks" section at the end of the file is a copy-paste of almost all the rules we already explained in
  detail. This wastes tokens and means if we want to change a rule later, we have to update it in two places.
  We should clean it up to only include quick, unique checks.
 4) Missing reviews for PCI drivers (Trigger Gap)
  The prompt only loads if the file is inside  drivers/pci/ . However, network, USB, or graphics drivers also use PCI
  functions. If a network driver has a bug in its PCI setup, this prompt won't load, and Sashiko will miss it. We should
  make the prompt load whenever common PCI functions are used anywhere in the codebase.

kwilczynski:
@rgushchin thank you for the feedback!

Some things that I didn't think about from the LLM angle. Let me incorporate some feedback from Bjorn, and then from you, and I will push revised document.

kwilczynski:
@rgushchin let me know if this is any better.

third_party/prompts/kernel/subsystem/pci.md
```diff
@@ -1,114 +1,696 @@
...
-dereferenced. For example, `pci_epf_destroy()` dereferences its argument
-unconditionally (to call `device_unregister()`), so passing an ERR_PTR
-value to it will crash. Note that `pci_epc_put()` is safe against ERR_PTR
-values because it checks `IS_ERR_OR_NULL()` before proceeding.
+Every finding has one of three dispositions:
 
-The following PCI endpoint functions return `ERR_PTR()` on failure, not NULL:
+- **REPORT as bugs**: code-correctness invariants; report as a bug.
+- **Low-severity suggestion**: prose, spelling, formatting, style.
+- **Silent**: unsure, or evidence insufficient to confirm.
```
rgushchin, line 16: It might actually hide some real issues. Not a hard blocker, just a note. It seems like there is some % of issues which AI can't reliably decide if it's real or not, so it just really depends if we prefer to miss some bugs or have less noise.

third_party/prompts/kernel/subsystem/pci.md
```diff
@@ -1,114 +1,696 @@
...
+When unsure whether a rule applies, or when the evidence needed to
+confirm a violation is not visible in the diff, stay silent.  Check
+the code-correctness
+invariants (the sections marked **REPORT as bugs**) before prose,
+spelling, and formatting items, so that under an output budget the
+bugs are reported first.  This file is a subsystem module loaded after
+`review-core.md`, which owns output format and Fixes-tag policy.
+The prose, spelling, and style sections below apply only when
+subjective review is enabled; render their findings per
+`inline-template.md`.  Every REPORT-as-bugs finding contributes at
```
rgushchin, line 27: This all mechanics doesn't belong to per-subsystem prompts, let's keep it separately.

third_party/prompts/kernel/subsystem/pci.md
```diff
@@ -1,114 +1,696 @@
...
+  "Endpoint" role.
+- Signal names carry the spec hash and capitalization: `PERST#`, `WAKE#`,
+  `CLKREQ#`.
+- Compound terms are one word, no hyphen or space: "Downstream", "Upstream",
+  "Retrain", "Rescan".
+- PCIe spec terms use the spec's capitalization ("Completer", "Requester
+  ID", "Message"), which doubles as a hint that they are defined in the
+  spec.
+
+## Referring to Functions, Structs, and Callbacks in Prose
```
rgushchin, line 682: I think it's pretty much the same across all subsystems. Do we really need it here? Do you see any specific problems?

rgushchin:
Thanks, Krzysztof! It's definitely better, but I still have some comments, so probably one more pass before I can merge it. Thanks for working on it!

rgushchin:
@kwilczynski is this still actual? It's in the draft state now, so Idk if you plan any further improvements.

kwilczynski:
@rgushchin yes. Sorry for the delay. This is still on my to do list.

PR #300, opened 2026-06-29
Support B4 Relay Original Author

### LINKED ISSUE #262: Insufficient CC-addresses - Sashiko does not reply to author using b4 relay
b4 relay uses additional email address for sending, except author's address:

> From: Aleksa Paunovic via B4 Relay <devnull+aleksa.paunovic.htecgroup.com ....
> To: ....
> Cc .... Aleksa Paunovic <aleksa.paunovic@

https://lore.kernel.org/all/20260513-riscv-time-mmio-v7-1-f8925ffc226b@htecgroup.com/

However Sashiko's reply https://lore.kernel.org/all/20260514055333.A29B8C2BCB7@smtp.kernel.org/ goes only to B4 Relay, not to the author. I think as result the author does not receive Sashiko's review email.

davidehu-69:
This PR enables Sashiko to correctly identify and parse the original author of patches submitted via B4 Relays. 

Currently, when a developer submits patches using a B4 Relay, the standard `From` header contains a mailing list proxy address (e.g. `Real Author via B4 Relay <devnull+author.example.com@kernel.org>`). Because Sashiko parses this as the patch author, the real developer never receives review notifications. This PR extracts the real developer's identity from the `X-Original-From` header.

Fixes #262

---

### Key Changes

*   Configured `MessageParser` to explicitly register and parse `X-Original-From` as an address header.
*   Updated `parse_email` in `src/patch.rs` to extract the real name and email address from `X-Original-From` if present, falling back to the standard `From` header.
*   **Verification Alignment:** Added a security check in `is_b4_alias` to verify that the `From` header is a valid B4 alias corresponding to the email in `X-Original-From`. We restrict this check to trusted B4 relay domains (`kernel.org` and `linux.dev`).
    *   *Note:* Scraping the production deployment confirmed that all of the 692 existing B4 relay submissions in the database are currently associated with the `kernel.org` domain.

---

## Discussion & Open Questions

1. **Do we need Commit `feat: Enable relaxed author matching for B4 Relay aliases during merge`?**
   * **Context:** If a developer is in the middle of sending a multi-part patch series when this update is deployed, their older patches will be stored under the `devnull+...` alias, but new incoming patches will be stored under their real email address.
   * **What the commit does:** It allows the database to match and merge them anyway, keeping the series in one piece.
   * **Question:** Is this transition support worth the extra matching code in the database layer, or are we okay with temporary series fragmentation during the deployment window to keep the code simpler?

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -2020,8 +2020,7 @@ impl Database {
             };
 
             // Relaxed author check logic
-            let author_match = crate::patch::extract_email(&existing_author)
-                == crate::patch::extract_email(author);
+            let author_match = crate::patch::authors_match(&existing_author, author);
             let series_match = (total_parts > 1 && total_parts == existing_total)
                 || existing_total == 1
                 || total_parts == 1;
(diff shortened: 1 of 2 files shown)


rgushchin:
Thanks!

PR #318, opened 2026-07-10
git_ops: make git_status independent of user config

ftabba:
`git_status` shells out to a bare `git status`, whose output wording
depends on the ambient `status.showUntrackedFiles` setting. On a host
where a user has set `status.showUntrackedFiles = no` in their global
config, a clean tree reports:

```
nothing to commit (use -u to show untracked files)
```

instead of:

```
nothing to commit, working tree clean
```

This makes `test_git_ops_extensions` fail on such hosts, since it
asserts on the "working tree clean" wording.

Pass `--untracked-files=normal` so the command's output is deterministic
regardless of the user's global git configuration.


rgushchin:
Thanks!

Issue #331, opened 2026-07-16
improve configuration to support per-subsystem branch selection


rgushchin:
Steve French <smfrench@gmail.com> writes:

    > The question came up recently when Sashiko couldn't review (failed to
    > apply) some ksmbd patches sent to the linux-cifs mailing list, what
    > branch does Sashiko apply the patches sent to the linux-cifs mailing
    > list?  mainline or cifs-2.6.git for-next (which linux-next pulls from)
    > or something else

    Sashiko determines patch baselines by matching modified file paths
    against MAINTAINERS T: entries (git://git.samba.org/ksmbd.git for
    ksmbd and git://git.samba.org/sfrench/cifs-2.6.git for cifs). Because
    neither T: line specifies an explicit branch name, Sashiko defaults to
    testing against the repository's default HEAD branch rather than
    for-next. If application against HEAD fails, it falls back to
    attempting linux-next and mainline master.

    I plan to add a configuration to Sashiko to make easier to customize.

derekbarbosa:
Hi Roman,

I just pushed a PR to update the docs for some existing subsystem mechanisms. We could try investigating in that direction as a start.

I am thinking that we could easily modify the existing subsystem config struct to reflect a base_tree/branch
```
   1 pub struct SubsystemMapping {                                                                                                                                                                                  
   2     pub pattern: String,                                                                                                                                                                                       
   3     pub name: String,                                                                                                                                                                                          
   4     // New fields:                                                                                                                                                                                             
   5     pub base_tree: Option<String>,                                                                                                                                                                             
   6     pub base_branch: Option<String>,                                                                                                                                                                           
   7 }         
```
to allow for

```
    1 [[subsystems.mapping]]
    2 pattern = ".*fs/smb/server/.*"
    3 name = "ksmbd"
    4 base_tree = "git://git.samba.org/ksmbd.git"
    5 base_branch = "for-next" # Overrides the HEAD default

```

please let me know if this thinking is misguided.

PR #332, opened 2026-07-16
docs: clarify [subsystems] configuration block

derekbarbosa:
The [subsystems] configuration block in Settings.toml applies to both NNTP (mailing lists) and Webhooks (Git forges). However, the documentation previously implied it was specific to the forge integration.

This commit updates the documentation to:
- Detail how the mapping works for both email addresses and file paths.
- Explain the default state (empty array with fallback heuristics).
- Move the subsystems documentation to a top-level section in the README and configuration reference.

Assisted-by: Gemini <noreply@google.com>


derekbarbosa:
triggered-by: #331

PR #333, opened 2026-07-16
index.html: fix copy to clipboard for inline reviews

### LINKED ISSUE #327: Copy-paste removes blank lines?
From Steven Rostedt:

[*] Roman, the paste from the copy icon appears to remove blank lines.
    This makes the output difficult to read as everything is bunched
    together. I typically go through and manually add back those blank
    lines. Is this a known issue? Can it be fixed?

    Of course, if I now have Sashiko replying to the tracing mailing
    list, I may not need to use that feature anymore.

derekbarbosa:
The copy button for inline reviews used innerText which dropped
newlines between the divs generated by escapeHtml.

Update copyToClipboard to handle HTMLElement input and manually join
the text content of child divs with newlines when a container is
passed.

Closes: #327

Assisted-by: Gemini <noreply@google.com>

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -1728,9 +1728,19 @@
             return full ? d.toLocaleString() : d.toISOString().split('T')[0];
         }
 
-        function copyToClipboard(text) {
+        function copyToClipboard(input) {
+            let text;
+            if (input instanceof HTMLElement) {
+                // If it's the div container, we want to preserve newlines from its child divs
+                // innerText usually does this, but sometimes it's inconsistent or lacks trailing newlines.
+                // For .log-inline-block which contains multiple <div> lines, we can join them.
+                const lines = Array.from(input.querySelectorAll('div')).map(d => d.innerText);
+                text = lines.length > 0 ? lines.join('\n') : input.innerText;
+            } else {
+                text = String(input);
+            }
             navigator.clipboard.writeText(text).then(() => {
-                console.log('Copied:', text);
+                console.log('Copied text');
             });
         }
 
@@ -2658,7 +2668,7 @@
                     <div style="margin-top:8px;">
                         <button class="toggle-btn" onclick="toggleCollapsible('inline-${uniqueId}', this)">${icon} Inline Review</button>
                         <div id="inline-${uniqueId}" class="collapsible ${isOpen}" style="position: relative;">
-                            <button class="copy-btn" style="position: absolute; top: 5px; right: 5px; z-index: 10;" onclick="copyToClipboard(this.nextElementSibling.innerText)">Copy</button>
+                            <button class="copy-btn" style="position: absolute; top: 5px; right: 5px; z-index: 10;" onclick="copyToClipboard(this.nextElementSibling)">Copy</button>
                             <div class="log-inline-block">${escapeHtml(r.inline_review, true, true)}</div>
                         </div>
                     </div>


derekbarbosa:
need to test with some production data first.

derekbarbosa:
Tested on a few patches, heres some sample output (from an older patch)
```
commit 089e5d253c0545fb289d65bfee21a9e2b0b4a271
Author: Wolfram Sang <wsa+renesas@sang-engineering.com>

hwspinlock: add callback to fill private data of a hwspinlock

This patch adds an init_priv callback to the hwspinlock_ops structure to allow
providers to initialize the private data of each hwspinlock. It updates the
hwspin_lock_register functions to accept an init_data pointer, passing it
along to the callback during lock registration.

> diff --git a/drivers/hwspinlock/hwspinlock_core.c b/drivers/hwspinlock/hwspinlock_core.c
> index 2c9eceba7fe8..afe1e7ce2829 100644
> --- a/drivers/hwspinlock/hwspinlock_core.c
> +++ b/drivers/hwspinlock/hwspinlock_core.c
[ ... ]
> @@ -538,6 +540,14 @@ int hwspin_lock_register(struct hwspinlock_device *bank, struct device *dev,
>  		spin_lock_init(&hwlock->lock);
>  		hwlock->bank = bank;
>  
> +		if (ops->init_priv) {
> +			hwlock->priv = ops->init_priv(i, init_data);
> +			if (IS_ERR(hwlock->priv)) {
> +				ret = PTR_ERR(hwlock->priv);
> +				goto reg_failed;

If init_priv() fails here and jumps to reg_failed, could this lead to a
use-after-free if a lock was concurrently requested?

Because hwspin_lock_register_single() inserts locks into the global
hwspinlock_tree one by one and drops the hwspinlock_tree_lock each time, a
concurrent consumer might acquire a newly registered lock before the whole bank
finishes registration.

If init_priv() fails on a later lock, the rollback loop in reg_failed calls
hwspin_lock_unregister_single() for the already acquired lock:

hwspin_lock_register() {
    ...
reg_failed:
	while (--i >= 0)
		hwspin_lock_unregister_single(base_id + i);
	return ret;
}

Since the lock's HWSPINLOCK_UNUSED tag was cleared by the consumer,
unregistration fails and returns NULL. The rollback loop ignores this failure
and returns an error, causing the bank array to be freed. This leaves the
consumer holding a valid handle pointing to freed memory.

> +			}
> +		}
> +
>  		ret = hwspin_lock_register_single(hwlock, base_id + i);

This isn't a bug introduced by this patch, but it appears
hwspin_lock_register_single() silently ignores radix_tree_insert() failures?

If radix_tree_insert() fails (e.g., due to -ENOMEM), it jumps to out:

hwspin_lock_register_single() {
	...
	ret = radix_tree_insert(&hwspinlock_tree, id, hwlock);
	if (ret) {
		if (ret == -EEXIST)
			pr_err("hwspinlock id %d already exists!\n", id);
		goto out;
	}
	...
out:
	mutex_unlock(&hwspinlock_tree_lock);
	return 0;
}

The registration seems successful to the caller since it returns 0.

During driver unload, hwspin_lock_unregister() iterates to remove locks. When
it calls hwspin_lock_unregister_single() on the failed lock, it returns NULL
because the lock is not in the tree:

hwspin_lock_unregister() {
	...
		tmp = hwspin_lock_unregister_single(bank->base_id + i);
		if (!tmp)
			return -EBUSY;
	...
}

Will this cause the unregistration loop to abort immediately and skip the
unregistration of all remaining successfully registered locks? Those remaining
locks stay in the global hwspinlock_tree, and the subsequent freeing of the
bank memory might lead to a use-after-free upon consumer access.

>  		if (ret)
>  			goto reg_failed;
>  	}

```

PR #339, opened 2026-07-19
media prompts: recognize V4L2 check_state callback invariant

### LINKED ISSUE #334: From Mauro Carvalho Chehab, media promtps improvements needed
Sashiko (as other static analyzers would likely do) got it
wrong about NULL checks at v4l2_subdev_state_get_format().

Basically, the media core has a complex mechanism to check
passed parameters before calling driver-specific callbacks.
This is done inside drivers/media/v4l2-core/v4l2-subdev.c,
on this function:

	static int check_state(struct v4l2_subdev *sd, struct v4l2_subdev_state *state,
                               u32 which, u32 pad, u32 stream)
	{
         	if (sd->flags & V4L2_SUBDEV_FL_STREAMS) {
        #if defined(CONFIG_VIDEO_V4L2_SUBDEV_API)
	                if (!v4l2_subdev_state_get_format(state, pad, stream))
				return -EINVAL;
                        return 0;
        #else
             		return -EINVAL;
	#endif
              	}

                if (stream != 0)
                        return -EINVAL;

		if (which == V4L2_SUBDEV_FORMAT_TRY && (!state || !state->pads))
                        return -EINVAL;

		return 0;
        }

The code Sashiko is complaining (csi_formatter_subdev_enum_mbus_code)
is called by a wrapper that ultimately calls:

	static int call_enum_mbus_code(struct v4l2_subdev *sd,
                                       struct v4l2_subdev_state *state,
                                       struct v4l2_subdev_mbus_code_enum *code)
	{
                if (!code)
                        return -EINVAL;

		return check_which(code->which) ? : check_pad(sd, code->pad) ? :
		       check_state(sd, state, code->which, code->pad, code->stream) ? :
	               sd->ops->pad->enum_mbus_code(sd, state, code);
	}

This should warrant that:
	fmt = v4l2_subdev_state_get_format()

would always return a non NULL pointer.

The same logic is there on other places. That's basically why
the code inside V4L2 subdev callbacks don't need to check
for NULL. We need to tell Sashiko somehow to not produce false
positives on such checks.

OllieinCanada:
Fixes #334

## Summary

Teach Sashiko to account for caller-side V4L2 subdevice state validation before
reporting a possible NULL dereference involving
`v4l2_subdev_state_get_format()` inside a registered driver callback.

The Media/V4L2 subsystem prompt requires the reviewer to inspect the actual
media-core callback path and dismiss the concern only when the matching wrapper
proves that the same state, pad, stream, operation, and relevant configuration
were validated before callback dispatch.

## Root cause

The affected driver callbacks are reached indirectly through V4L2 operation
tables. Their validating wrappers live in the media core rather than in the
driver callback itself.

Reviewing only the callback can therefore make a protected lookup appear to be
an unchecked dereference.

## Changes

- add a Media/V4L2 subsystem prompt
- register Media/V4L2 paths and symbols in the subsystem prompt index
- require concrete caller-side proof before dismissing the warning
- retain the warning when a caller bypasses the wrapper, values differ, the
  guard is conditional, or the required context is unavailable

## Scope

This change does not assume that
`v4l2_subdev_state_get_format()` is globally non-NULL and does not suppress
V4L2 NULL-dereference findings broadly.

Following maintainer feedback, the submitted diff is intentionally limited to
the subsystem prompt and its index entry.

## Validation

- confirmed the cumulative pull-request diff contains only the Media/V4L2
  prompt and subsystem index
- manually inspected the prompt’s guarded and unsafe reasoning boundaries
- `git diff --check` passes
- no model-backed regression tests are included in this pull request


rgushchin:
Thank you for the PR! Prompt changes look good to me, but I'd drop the rest. I don't think we want to add tests like this right now - they might not be stable due to the nature of LLMs. We can return to this question separately. Also, please don't change the REVISION file. Let's just change the prompt and the index file. Thanks!

OllieinCanada:
Thanks for the review. I’ve narrowed the PR to the Media/V4L2 prompt and subsystem index only, and removed the tests, fixtures, prompt-bundle changes, README update, and REVISION change. The cumulative diff now contains only those two files.

PR #338, opened 2026-07-19
[RFC] Add support for Ollama provider

mchehab:
The OpenAI compatible model doesn't work with Ollama. Add another provider for it.

Assisted-by: qwen2.6_claude_27B

PR #359, opened 2026-07-22
reviewer: limit max concurrency semaphore

mchehab:
The reviewer code works with two semaphores:
- semaphore, which reflects what's there at concurrency;
- llm_semaphore, which is initialized to have 3 times the value of concurrency.

This is based on an estimation that an active patch review consumes ~3 LLM slots.

Such model works when using cloud-based models, but when the models run locally, this must be aligned with the maximum queue size of the local engine.

Add an extra optional parameter to allow adjusting such limit.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -159,6 +159,7 @@ Optional array of additional git remotes to track.
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | `concurrency` | integer | -- | Number of concurrent reviews. |
+| `max_concurrency` | integer | -- | Max number of concurrent review queue. |
 | `worktree_dir` | string | -- | Directory for git worktrees used during review. |
 | `timeout_seconds` | integer | `3600` | Maximum time per review (seconds). |
 | `max_retries` | integer | `3` | Retry count on transient failures. |
--- a/src/reviewer.rs
+++ b/src/reviewer.rs
@@ -96,7 +96,8 @@ impl Reviewer {
     /// * `db` - The database connection.
     /// * `settings` - Application settings.
     pub async fn new(db: Arc<Database>, settings: Settings) -> Self {
-        let concurrency = settings.review.concurrency;
+        let mut concurrency = settings.review.concurrency;
+        let max_concurrency = settings.review.max_concurrency;
         let repo_path = PathBuf::from(&settings.git.repository_path);
 
         let baseline_registry =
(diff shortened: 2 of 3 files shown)


mchehab:
@derekbarbosa This is per your request at #338: even after having it merged, it still makes sense to be able to optionally limit `llm_concurrency` when `concurrency` > 1.

derekbarbosa:
Hi Mauro, 

Sorry that this hasn't been merged yet. 

Would you mind implementing the following feedback? A few things appear to have drifted in the past month.

🔴 Safety / Bugs
1. max_concurrency = 0 deadlocks the entire reviewer.
Both concurrency and llm_concurrency are reduced via .min(max). A value of 0 produces Semaphore::new(0), which grants no permits. Every acquire_owned().await then blocks forever — the reviewer silently stops processing any patchset. The existing code already guards llm_concurrency with max(1, …), but the new cap bypasses that guard.

```
-        if max_concurrency.is_some() {
-            concurrency = concurrency.min(max_concurrency.unwrap_or(usize::MAX));
-            llm_concurrency = llm_concurrency.min(max_concurrency.unwrap_or(usize::MAX));
-        }
+        if let Some(max) = max_concurrency {
+            concurrency = concurrency.min(max).max(1);
+            llm_concurrency = llm_concurrency.min(max).max(1);
+        }
```
(This also removes the dead unwrap_or(usize::MAX) — the MAX branch is unreachable inside an is_some() guard.)

🟠 Regressions
2. One value is applied to two quantities with a 3:1 relationship, so capping the LLM queue also silently throttles review parallelism.
concurrency and llm_concurrency are semantically different (llm_concurrency = 3 × concurrency). The commit message states the intent is to align the LLM queue with the local engine's queue size, but the code also caps concurrency 
(worktrees/processes) by the same absolute value. With concurrency = 16, max_concurrency = 4:

```
- before: concurrency = 16, llm_concurrency = 48
- after: concurrency = 4, llm_concurrency = 4
```
Four concurrent reviews now share 4 LLM slots (1 each) instead of 12, contradicting the documented "~3 LLM slots per review" model and under-saturating the LLM. This needs a decision (per GEMINI.md: "If not sure, ask the user"):

```
-        if let Some(max) = max_concurrency {
-            concurrency = concurrency.min(max).max(1);
-            llm_concurrency = llm_concurrency.min(max).max(1);
-        }
+        // Cap only the LLM request queue; leave review parallelism alone.
+        if let Some(max) = max_concurrency {
+            llm_concurrency = llm_concurrency.min(max).max(1);
+        }
```
Alternatively, if max_concurrency is meant to cap reviews, it should scale the LLM cap (llm_concurrency.min(max * 3)) rather than reusing the raw value. Either way, one knob should not cap two unrelated quantities.

🔵 Style / Nits
3. Startup log now reports the raw, uncapped value.
start() logs self.settings.review.concurrency (e.g. 16), while the semaphore was capped to 4 in new(). The log is misleading after this change. Log the effective value in new():

```
         if let Some(max) = max_concurrency {
             concurrency = concurrency.min(max).max(1);
             llm_concurrency = llm_concurrency.min(max).max(1);
         }
 
+        info!(
+            "Reviewer concurrency: {} (LLM slots: {})",
+            concurrency, llm_concurrency
+        );
+
 ```
4. LocalReviewReviewSettings.max_concurrency is dead in the CLI path.
src/local_review.rs (run_worker, lines 275-290) reads only concurrency and feeds it to buffer_unordered(concurrency) (line 824). The new max_concurrency field is never consumed, so setting it in a local Settings.toml has no effect. Either wire it into the local-review path or drop the field to avoid implying support that isn't there.

5. No test for the new capping logic (GEMINI.md §5 requires unit tests for new logic). A small test asserting the effective concurrency/llm_concurrency for a few (concurrency, max_concurrency) pairs would lock in the semantics.

PR #368, opened 2026-07-24
git: make bad-tag recovery test deterministic

### LINKED ISSUE #350: tests: make bad-tag recovery test portable across Git versions
## Summary

`git_ops::tests::test_ensure_remote_bad_tag_recovery` fails consistently with
Git 2.25.1 because `git fetch --prune --no-tags` succeeds without inspecting or
removing the deliberately corrupt local tag.

The recovery logic is entered only when fetch fails with stderr containing
`fatal: bad object refs/tags/`. On this Git version that diagnostic is not
produced, so `ensure_remote()` returns successfully while
`.git/refs/tags/bad-tag` remains and the final assertion fails.

## Reproduction

On current `main` at `8ac8e51fd79c8560404efe62d0c9ce38defce5af`:

```text
cargo test --release --lib \
  git_ops::tests::test_ensure_remote_bad_tag_recovery -- --exact
```

Result:

```text
test git_ops::tests::test_ensure_remote_bad_tag_recovery ... FAILED

Bad tag should have been deleted by recovery logic
```

The failure also causes `make check-pr` to fail locally.

## Environment

- Rust 1.90.0
- Git 2.25.1
- WSL Ubuntu

## Expected behavior

The test should deterministically exercise the recovery path across supported
Git versions, or the repository should document and enforce a minimum Git
version that produces the expected fetch failure.

This was discovered while validating an unrelated NNTP whitespace fix. That
change does not modify `src/git_ops.rs`, and the failure reproduces when this
test is run alone.

OllieinCanada:
Fixes #350

## Summary

Make the bad-tag recovery regression test deterministic across Git versions.

## Root cause

The previous test depended on `git fetch --prune --no-tags` failing after a
corrupt local tag was created. Git 2.25.1 completes that fetch successfully
without inspecting or removing the tag, so Sashiko's recovery branch was never
entered and the assertion failed.

## Changes

- extract the existing bad-tag deletion step into a private helper
- keep the production fetch and retry behavior unchanged
- invoke the recovery helper with the precise Git diagnostic in the test
- remove the unnecessary remote repository and version-dependent fetch setup

## Validation

- reproduced the original failure with Git 2.25.1
- focused bad-tag recovery test passes
- all six `git_ops` tests pass
- `cargo clippy --all-targets --all-features --release -- -D warnings`
- `cargo test --release`
- `git diff --check`
- `make check-pr`

### DIFF
--- a/src/git_ops.rs
+++ b/src/git_ops.rs
@@ -510,6 +510,32 @@ fn get_worktree_lock() -> Arc<AsyncMutex<()>> {
         .clone()
 }
 
+async fn recover_bad_tag(repo_path: &Path, stderr: &str) -> bool {
+    if !stderr.contains("fatal: bad object refs/tags/") {
+        return false;
+    }
+
+    let Some(start) = stderr.find("refs/tags/") else {
+        return false;
+    };
+    let tag_path = stderr[start..].split_whitespace().next().unwrap_or("");
+    if tag_path.is_empty() {
+        return false;
+    }
+
+    warn!(
+        "Detected bad tag '{}'. Attempting to delete and retry fetch.",
+        tag_path
+    );
+    let _ = Command::new("git")
+        .current_dir(repo_path)
+        .args(["update-ref", "-d", tag_path])
+        .output()
+        .await;
+
+    true
+}
+
 pub async fn ensure_remote(
     repo_path: &Path,
     name: &str,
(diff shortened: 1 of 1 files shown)


rgushchin:
Closing this PR because the underlying issue was already organically solved by adding `--no-tags` to the fetch command in May. I've pruned the obsolete recovery loop entirely in the latest push.

PR #369, opened 2026-07-24
reviewer: skip patches meant for xfstests

### LINKED ISSUE #103: Don't try to review patches meant for xfstests
It is quite common for patches against xfstests to be cc'ed to kernel mailing lists, such as ext4.  For example[1].  

[1] https://sashiko.dev/#/patchset/20260418223850.20351-1-ablagodarenko%40ddn.com

Since new tests involve creating new files, that means that Sashiko can mistakingly believe that the patches "apply" against the kernel, resulting in the Sashiko review:

> This patch appears to add an xfstests script directly to the Linux kernel
> repository. Should this be submitted to the xfstests-dev repository instead?
> Applying this to the kernel tree creates unusable test code, as the common
> scripts it sources do not exist in the Linux tree.

Sashiko did seem to understand that the patch should be sent to the xfstests-dev repository.  But that was actually the intent!

Finally, there is an opportunity to save LLM token budget burns by skipping any patches that appear to be meant for xfstests.   :-)

OllieinCanada:
Fixes #103

## Summary

Skip patches that are clearly intended for xfstests before Sashiko attempts
Linux baseline selection, patch application, or AI review.

## Root cause

New xfstests cases commonly add files below `tests/<suite>/<case>`. Because new
files can apply cleanly to the Linux checkout, Sashiko can mistake the patch
for a kernel change and spend model tokens reviewing it in the wrong
repository.

## Changes

- recognize diffs whose changed files are exclusively in the top-level
  `tests/<suite>/<path>` layout used by xfstests
- mark those patches as skipped before baseline resolution or review work
- complete an all-xfstests patchset without creating an AI review record
- keep Linux selftests under `tools/testing/`, normal kernel paths, empty
  diffs, and mixed kernel/xfstests diffs in the existing review path

## Validation

- focused positive and negative path-classification tests pass
- pipeline test confirms the patch becomes `Skipped`, the patchset becomes
  `Reviewed`, and no review record is created
- maintainer-requested path-check simplification passes its focused test
- GitHub Actions DCO, full lint, and Rust unit-test checks pass
- `git diff --check`


src/reviewer.rs
```diff
@@ -71,6 +71,18 @@ fn generate_id() -> String {
...
+    let files = extract_files_from_diff(diff);
+
+    !files.is_empty()
+        && files.iter().all(|file| {
+            let mut components = file.split('/');
+            components.next() == Some("tests")
+                && components.next().is_some_and(|part| !part.is_empty())
+                && components.next().is_some_and(|part| !part.is_empty())
+        })
+}
```
derekbarbosa, line 92: a minor nitpick, but you could simplify  `.next()` and the long RHS comparison with:
```
    !files.is_empty() && files.iter().all(|file| {
        let mut parts = file.split('/').filter(|p| !p.is_empty());
        parts.next() == Some("tests") && parts.count() >= 2
    })
}
```

derekbarbosa (commented):
Hi, sorry for the delay!

Thank you for the fix. This looks good. The review comments I posted are minor fixups.

However, this has me thinking if we should try and come up with a solution that would allow certain filepaths to be ignored per-subsystem.

```
        // Linux selftests live under tools/testing. Top-level tests/<suite>/<case>
        // paths identify xfstests patches, which must not be reviewed as kernel code.
```

but I am unsure if xfstests' usecase is unique or not. don't consider this a blocker for now :)

src/reviewer.rs
OllieinCanada: Addressed in 47446f9. Thanks—the classifier now uses the simpler iterator/count form you suggested, with behavior unchanged. The focused classifier test passes.

OllieinCanada:
Thanks for the review. I’ve pushed the minor path-check simplification and resolved that thread. I agree that configurable per-subsystem ignored paths may be useful beyond xfstests; since it is not a blocker, I’ve kept this PR narrowly scoped to #103 rather than expanding it here.

OllieinCanada:
Refreshed this branch against current main and resolved the reviewer test-module overlap while preserving both the xfstests coverage and upstream interaction-ID coverage. The PR is conflict-free and DCO, lint, and unit tests are all green.

OllieinCanada:
Refreshed the branch to current main in signed commit e664baf. The cumulative diff remains `src/reviewer.rs` only. In a WSL-native exact-commit checkout, all 13 reviewer tests pass (including the positive, negative, and no-AI xfstests cases), formatting passes, all-features clippy passes with warnings denied, cumulative `git diff --check` passes, and DCO passes.

derekbarbosa:
from `review-pr` skill:

🔴 Incomplete classification — real xfstests patches touching common/, src/, m4/ are not skipped
is_xfstests_patch requires every file to live under tests/<suite>/<case>. But fstests/xfstests patches routinely modify shared helpers outside that tree (common/rc, common/config, src/*.c, m4/*, check, Makefile), or mix a new tests/ext4/065 with a common/rc change. All of those fail the .all(...) test and flow straight into kernel review, where baseline apply fails or the AI reviews a shell script as C. The fix is only partial.

```
 fn is_xfstests_patch(diff: &str) -> bool {
+    // fstests/xfstests top-level dirs that do not exist in the kernel tree.
+    const XFSTESTS_ROOTS: [&str; 4] = ["tests", "common", "src", "m4"];
+
     let files = extract_files_from_diff(diff);
-
     !files.is_empty()
         && files.iter().all(|file| {
-            let mut components = file.split('/').filter(|part| !part.is_empty());
-            components.next() == Some("tests") && components.count() >= 2
+            file.split('/')
+                .find(|part| !part.is_empty())
+                .is_some_and(|root| XFSTESTS_ROOTS.contains(&root))
         })
 }
```
🟠 Mixed patchsets drop xfstests patches from the LLM payload — patch index becomes non-contiguous
In a mixed series (e.g. patch 1 = fs/ext4/inode.c, patch 2 = tests/ext4/065, patch 3 = fs/ext4/namei.c), the xfstests patch is partitioned out, but the remaining patches keep their original index values (1, 3). input_payload["patches"] is sent to the LLM with indices 1 and 3 and no indication that a patch was dropped, so the model sees "patch 3" with a payload of two patches. This mirrors pre-existing subject-filtering behavior, but that is user-opt-in while this is automatic, so it will fire in the middle of real series more often. Consider preserving the original 1-based index (matching the subject/mail header) and, if desired, annotate skipped indices or pass metadata indicating which patches were omitted/skipped.

🟡 Auto-skip records no reason in the DB (observability)
update_patch_status(*patch_id, ReviewStatus::Skipped.as_str()) writes a bare Skipped with no result_description, unlike the "touches only ignored files" and --no-ai paths (reviewer.rs:1231, reviewer.rs:1488) which also persist a reviews row explaining the skip. A Skipped patchset/patch now has no way to distinguish "xfstests" from a user --skip-subject filter in reporting (the SQL at db.rs:3574/db.rs:3590 already discriminates on result_description). Log-only provenance is lost once the log rotates.

🟢 Design alignment — no design doc, and the rule is a magic constant
GEMINI.md requires a design document for non-trivial features, and this is a policy decision (which paths are silently never reviewed). The tests/<suite>/<case> convention and the 4-component depth rule are hard-coded with no config surface, unlike skip_filters/only_filters which are user-configurable. Aligning this with subject-filtering.md (a declarative filter rule instead of an inlined heuristic) would make it reviewable and testable against new fstests layouts. No action-blocking issue, but worth a follow-up design note.

OllieinCanada:
Addressed the review in `243bc10`:

- xfstests-only classification now recognizes the project’s shared `common/`, `src/`, and `m4/` roots in addition to `tests/`
- skipped patches now receive a durable skipped review row with `Skipped: patch contains only xfstests files`, so operators can tell policy from failure
- added regression coverage for every recognized root and for the persisted skip reason

Patch indices remain the original database indices intentionally; gaps show that an omitted patch existed, while compacting them would misidentify later patches in the source series.

Exact-head CI is green: DCO, lint, and the full Rust test job all pass.

PR #372, opened 2026-07-24
cli: support local review of root commits

### LINKED ISSUE #203: Issue reported by sashiko-cli when start review for local repo
cargo run --bin sashiko-cli local -r <repo path> <commit id>
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.21s
     9971617d8ec95fdccc8a18f71ae4b0f99fca7573 --force-local`
[1/4] Extracting patches from 9971617d8ec95fdccc8a18f71ae4b0f99fca7573... (1 commit)
[2/4] Starting review subprocess...
Local Review Results
  Input:    9971617d8ec95fdccc8a18f71ae4b0f99fca7573
  Baseline: 9971617d8ec95fdccc8a18f71ae4b0f99fca7573^

Error: Failed to resolve 9971617d8ec95fdccc8a18f71ae4b0f99fca7573^: fatal: ambiguous argument '9971617d8ec95fdccc8a18f71ae4b0f99fca7573^': unknown revision or path not in the working tree.

OllieinCanada:
Fixes #203

## Summary

Allow `sashiko-cli local` to review a repository's root commit.

Root commits are checked out and reviewed directly because they have no parent
baseline. Ordinary commits continue through the existing parent-based patch
application path.

## Root cause

The local CLI always constructed `<first commit>^` when no baseline was given.
That revision is invalid when the requested commit is the repository's first
commit, so the review subprocess exited before producing a result.

## Changes

- inspect the first commit's parent list before selecting the default baseline
- use the direct-commit path for a parentless root commit
- preserve root-commit file discovery and commit-message context for review
- add end-to-end CLI coverage for root and ordinary commits

## Validation

- root-commit CLI regression passes through the real review subprocess
- ordinary-commit CLI regression preserves the existing application path
- full release suite: 464 library, 16 main, 4 CLI, and 21 ordinary integration
  tests passed; expected ignored tests remained ignored
- `cargo clippy --all-targets --all-features --release -- -D warnings`
- cumulative `git diff --check`
- DCO sign-off check

No external Git server, AI provider, or paid API is required by the regression
tests.


derekbarbosa:
Hello,

> Allow sashiko-cli local to review a repository's root commit.
> 
> Root commits are checked out and reviewed directly because they have no parent
> baseline. Ordinary commits continue through the existing parent-based patch
> application path.

would you mind elaborating on this? I am having a bit of trouble understanding the perceived usecase here

OllieinCanada:
Sure. The use case is a local repository where the requested review starts with the repository's initial (root) commit—for example, a small reproducer repository, a newly initialized repository, or imported history whose first commit also needs review. Today, when no baseline is supplied, sashiko-cli constructs <commit>^ as the default baseline. A root commit has no parent, so Git cannot resolve that revision and the review exits before analyzing the commit. This PR handles only that boundary case by reviewing the root commit's tree directly. Commits that have a parent continue through the existing parent-baseline and patch-application path unchanged. The failure shown in #203 is consistent with this root-commit case, though confirmation from the reporter that the target SHA is the repository's root commit would remove the remaining uncertainty.

derekbarbosa:
> Sure. The use case is a local repository where the requested review starts with the repository's initial (root) commit—for example, a small reproducer repository, a newly initialized repository, or imported history whose first commit also needs review. Today, when no baseline is supplied, sashiko-cli constructs ^ as the default baseline. A root commit has no parent, so Git cannot resolve that revision and the review exits before analyzing the commit. This PR handles only that boundary case by reviewing the root commit's tree directly. Commits that have a parent continue through the existing parent-baseline and patch-application path unchanged. The failure shown in #203 is consistent with this root-commit case, though confirmation from the reporter that the target SHA is the repository's root commit would remove the remaining uncertainty.

ah. I see. I was struggling to understand the application. as in "who would be running this on a repo with virtually no git history", but your investigation suggests that this is useful for shallow clones (--depth=1) of kernel trees?

derekbarbosa:
pasting a review-pr run:

🔴 Safety / Bug — shallow clones are misdetected as root commits

local_review_target in src/bin/sashiko-cli.rs:1427 treats a commit with no parent listed by git rev-list --parents as a root commit. But that condition is also true for the boundary commit of a shallow clone (--depth=N), whose parent is simply absent from the object graph.

# this repo is itself a shallow clone:
$ git rev-list --parents -n 1 HEAD

42c3b83a 5afe9088 9db003da   # HEAD is a merge, fine — but:

$ git rev-list --parents -n 1 <shallow-boundary>   # returns just "<sha>", no parent

Why it's a regression:  the function conflates "commit has no parent at all" (true root) with "parent isn't present in this shallow object graph" (shallow boundary). The sashiko repo itself has a .git/shallow file, so this path is reachable in normal use. 

For a shallow boundary commit the new code returns (first_sha, Some(first_sha)), so the CLI passes --review-commit <sha> + --review-patch-index 1 and review_single_patch runs git diff-tree --root <sha>. --root then emits the entire tree as a giant synthetic "new file" diff rather than the actual patch. Previously this scenario failed loudly (git worktree add <sha>^ → clear error); now it silently produces a misleading full-tree review. 

Same trap exists in src/local_review.rs:458 — adding --root only because a commit looks root will blow up the file list 
for shallow boundaries.

Fix — distinguish a true root commit from a shallow boundary. A commit is only a root if git rev-list reports no parent and the repo has no shallow grafts:
```
async fn local_review_target(
    repo_path: &std::path::Path,
    first_sha: &str,
    baseline: Option<&str>,
) -> Result<(String, Option<String>)> {
    if let Some(baseline) = baseline {
        return Ok((baseline.to_string(), None));
    }

    let shallow = tokio::process::Command::new("git")
        .current_dir(repo_path)
        .args(["rev-parse", "--is-shallow-repository"])
        .output()
        .await
        .map(|o| o.status.success() && String::from_utf8_lossy(&o.stdout).trim() == "true")
        .unwrap_or(false);

    let output = tokio::process::Command::new("git")
        .current_dir(repo_path)
        .args(["rev-list", "--parents", "-n", "1", first_sha])
        .output()
        .await?;

    if !output.status.success() {
        return Err(anyhow::anyhow!(
            "Failed to inspect parents of {}: {}",
            first_sha,
            String::from_utf8_lossy(&output.stderr).trim()
        ));
    }

    let parents = String::from_utf8_lossy(&output.stdout);
    let mut revisions = parents.split_whitespace();
    revisions
        .next()
        .ok_or_else(|| anyhow::anyhow!("Git returned no commit for {}", first_sha))?;

    if revisions.next().is_some() {
        Ok((format!("{}^", first_sha), None))
    } else if shallow {
        // Shallow boundary, not a true root commit. Review its parent diff
        // normally; git will materialize the missing parent on demand.
        Ok((format!("{}^", first_sha), None))
    } else {
        // A true root commit has no baseline to apply against. Check it out
        // and review the commit directly instead.
        Ok((first_sha.to_string(), Some(first_sha.to_string())))
    }
}
```

🟡 Complexity / Logic — --root in review_single_patch is too broad

src/local_review.rs:458 now unconditionally passes --root to diff-tree. --root only changes output for commits that look root, so this is mostly harmless for real patches, but it makes the file-list heuristic dependent on the same shallow-vs-root ambiguity as above. If you fix the shallow detection in the CLI, also gate --root there on the same "true root" signal (or pass an explicit flag through WorkerOptions instead of inferring from diff-tree). Tying both call sites to one helper prevents the two from drifting.

🔵 Style / Test fragility — get_bin_path fallback prefers stale release/

tests/integration/cli_tests.rs:21 checks release/ before debug/. Under cargo test, CARGO_BIN_EXE_sashiko-cli is set correctly (verified: cargo preserves the hyphen), so the fallback only runs outside cargo. But when it does run and a stale target/release/sashiko-cli exists (as it does here — predating this PR), the test silently tests the wrong binary. 

Prefer the debug binary the current build just produced:
```
-    let release_path = path.join("release").join(&executable);
-    let debug_path = path.join("debug").join(&executable);
-
-    if release_path.exists() {
-        release_path
-    } else if debug_path.exists() {
-        debug_path
+    let debug_path = path.join("debug").join(&executable);
+    let release_path = path.join("release").join(&executable);
+
+    if debug_path.exists() {
+        debug_path
+    } else if release_path.exists() {
+        release_path
     } else {
         panic!("Could not find {name} binary in target/release or target/debug. Path: {path:?}");
     }
```

🟢 Design Alignment

The root-commit flow itself (when a commit is genuinely root) is correct and matches DESIGN_LOCAL_REVIEW_UX.md intent: local_review_target → --review-commit/--review-patch-index 1 → run_worker_in_worktree review_commit branch → patch_shas populated → diff-tree --root lists the initial tree. The new patch_messages capture in run_worker_in_worktree (src/local_review.rs:661) mirrors the existing patch_shows pattern and is safe.

PR #373, opened 2026-07-24
baseline: support per-subsystem branch configuration

### LINKED ISSUE #331: improve configuration to support per-subsystem branch selection

OllieinCanada:
Fixes #331

## Summary

Allow subsystem mappings to select the Git tree and branch used when Sashiko
tries to apply a patch series.

For example, ksmbd can now be configured to try its `for-next` branch:

```toml
[[subsystems.mapping]]
pattern = "^fs/smb/server/.*"
name = "ksmbd"
base_tree = "git://git.samba.org/ksmbd.git"
base_branch = "for-next"
```

## Root cause

Sashiko inferred subsystem trees from `MAINTAINERS`, but a `T:` entry without
an explicit branch caused it to try the remote's default HEAD. Some subsystem
patches are based on a development branch such as `for-next`, so they could
fail to apply even when the correct tree had been identified.

## Changes

- add optional `base_tree` and `base_branch` fields to subsystem mappings
- match configured baselines against the changed file paths
- try matching configured baselines before MAINTAINERS-derived trees
- retain explicit patch baselines as the highest-priority candidates
- preserve existing behavior for mappings without `base_tree`
- document the new configuration with a ksmbd example

## Validation

- synced with current upstream `main` and resolved baseline-priority changes
- focused settings and baseline-resolution tests pass
- `cargo clippy --release --lib -- -D warnings`
- GitHub Actions DCO, full lint, and Rust unit-test checks pass
- `git diff --check`

### DIFF
--- a/Settings.toml
+++ b/Settings.toml
@@ -23,10 +23,11 @@ enabled = false
 # webhook_secret = ""
 
 [subsystems]
-# Regex map: email address pattern -> subsystem name
+# Regex map: email address or changed path pattern -> subsystem name
 # mapping = [
 #     { pattern = ".*linux-kernel@vger.kernel.org.*", name = "LKML" },
 #     { pattern = ".*netdev@vger.kernel.org.*", name = "netdev" },
+#     { pattern = "^fs/smb/server/.*", name = "ksmbd", base_tree = "git://git.samba.org/ksmbd.git", base_branch = "for-next" },
 # ]
 
 [database]
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -179,13 +179,18 @@ This feature is **globally active** and applies to both mailing list (NNTP) and 
 |-----|------|---------|-------------|
 | `mapping` | list of objects | `[]` | A list of rules mapping a regular expression pattern to a subsystem name. |
 
-Each mapping object in the list requires two fields:
+Each mapping object requires two fields:
 * `pattern` (string): A regular expression used for matching.
 * `name` (string): The resulting subsystem name if the pattern matches.
+
+It also accepts two optional baseline fields:
+* `base_tree` (string): Git repository URL to try before MAINTAINERS-derived baselines when the pattern matches a changed file path.
+* `base_branch` (string): Branch in `base_tree` to use. When omitted, the remote's default HEAD is used.
 
 **How it works:**
 - **For Git Forges (Webhooks):** The system applies the `pattern` against the **file paths** modified by a pull request (e.g., matching `^drivers/net/.*`).
 - **For Mailing Lists (NNTP):** The system applies the `pattern` against the **To and Cc email addresses** of the incoming patch email.
+- **For Baseline Selection:** Mappings with `base_tree` are also matched against changed file paths. A matching tree and optional branch are tried before baselines inferred from `MAINTAINERS`, linux-next, and mainline.
 
 When a patch is tagged with a subsystem, it can trigger subsystem-specific AI review rules (context loading) and specific email/embargo policies defined in `email_policy.toml`.
 
@@ -196,6 +201,7 @@ mapping = [
     { pattern = ".*net/.*", name = "Networking" },
     { pattern = ".*fs/.*", name = "Filesystems" },
     { pattern = ".*mm/.*", name = "Memory Management" },
+    { pattern = "^fs/smb/server/.*", name = "ksmbd", base_tree = "git://git.samba.org/ksmbd.git", base_branch = "for-next" },
 ]
 ```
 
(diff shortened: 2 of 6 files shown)


derekbarbosa:
review-pr output:
```
🟠 Regressions — precedence vs custom remotes is undocumented
resolve_configured_subsystem_baselines is inserted at position 2, ahead of custom remotes (position 4). The docs claim baselines are tried "before MAINTAINERS, linux-next, and mainline" but omit custom remotes, which always emit candidates for every patch. A user with both a path-matching base_tree mapping and custom remotes will now see the config baseline win — a real, if unlikely, behavior change.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -215,7 +215,8 @@ mapping = [
 **How it works:**
 - **For Git Forges (Webhooks):** The system applies the `pattern` against the **file paths** modified by a pull request (e.g., matching `^drivers/net/.*`).
 - **For Mailing Lists (NNTP):** The system applies the `pattern` against the **To and Cc email addresses** of the incoming patch email.
-- **For Baseline Selection:** Mappings with `base_tree` are also matched against changed file paths. A matching tree and optional branch are tried before baselines inferred from `MAINTAINERS`, linux-next, and mainline.
+- **For Baseline Selection:** Mappings with `base_tree` are also matched against changed file paths. A matching tree and optional branch are tried before baselines inferred from `MAINTAINERS`, custom remotes, linux-next, and mainline.
🟡 Complexity / SOLID — pattern field is overloaded, and regex compilation is triplicated
The same pattern is now matched against three domains: emails (main.rs:2228), paths (main.rs:2280), and paths again for baselines (baseline.rs:387). An email-shaped pattern (e.g. .*netdev@vger.kernel.org.*) with base_tree can never match a path, silently never contributing a baseline. This is the third near-identical "compile mapping regexes" block. Recommend a single shared helper and a doc comment flagging the path-only baseline matching domain.
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -374,6 +374,9 @@ impl BaselineRegistry {
     fn resolve_configured_subsystem_baselines(
         &self,
         files: &[String],
         subsystem_mapping: &[SubsystemMapping],
     ) -> Vec<BaselineResolution> {
+        // NOTE: matches against file paths only, unlike the NNTP path which
+        // matches the same `pattern` against email addresses. A mapping whose
+        // `pattern` targets emails will never match here.
         let lower_files: Vec<String> = files.iter().map(|file| file.to_lowercase()).collect();
🔵 Style / Nits
1. resolve_candidates now takes a 4th positional parameter; every test call site was rewritten with &[]. Grouping into a BaselineResolutionConfig struct would avoid future call-site churn, but this is internal-only and acceptable.
2. No test covers base_branch: None (default-HEAD path). Add one:
--- a/src/baseline.rs
+++ b/src/baseline.rs
@@ -726,6 +726,27 @@ mod tests {
         assert!(matches!(
             &candidates[0],
             BaselineResolution::RemoteTarget { name, .. } if name == "net-next"
         ));
     }
+
+    #[tokio::test]
+    async fn test_configured_subsystem_baseline_without_branch_uses_head() {
+        let registry = create_registry();
+        let files = vec!["net/core.c".to_string()];
+        let mappings = vec![SubsystemMapping {
+            pattern: "^net/.*".to_string(),
+            name: "networking".to_string(),
+            base_tree: Some("git://example.com/networking.git".to_string()),
+            base_branch: None,
+        }];
+
+        let candidates = registry
+            .resolve_candidates(&files, "Subject", None, &mappings)
+            .await;
+
+        assert!(matches!(
+            &candidates[0],
+            BaselineResolution::RemoteTarget { branch, .. } if branch.is_none()
+        ));
+    }
3. Case handling: files are lowercased but the pattern is not — consistent with identify_subsystems_from_paths (main.rs:2287), so an uppercase pattern silently never matches. Existing convention; note only.
```

PR #382, opened 2026-07-30
Track linux-fscrypt and fsverity mailing lists

ebiggers:
Add linux-fscrypt and fsverity to the tracked mailing lists in the Kubernetes deployment configuration to enable monitoring of these lists.


rgushchin:
Merged, thank you!

Issue #49, opened 2026-03-25
Tree selection: support pending dependences


matttbe:
Hello,

It would be nice if there could be a way to mark pending patches dependencies.

I know two different techniques using dedicated tags in the cover-letter (or single patch, in the comment section):
- Using the `Based-on: <MESSAGE_ID>` tag: [Patchew supports that](https://lists.gnu.org/archive/html/qemu-devel/2017-08/msg01288.html)
- Using [b4's dependencies](https://b4.docs.kernel.org/en/latest/contributor/prep.html#working-with-series-dependencies) tags: `prerequisite-change-id`, `prerequisite-patch-id` and `prerequisite-message-id`.

dlech:
I've had several series recently "Fail to apply" because of this.

- https://sashiko.dev/#/patchset/20260714-iio-adc-ti-ads112c14-buffered-read-v1-0-fe6e1c971288%40baylibre.com
- https://sashiko.dev/#/patchset/20260724-iio-adc-ti-ads112c14-continuous-mode-v1-0-9eb0b7a4f020%40baylibre.com
- https://sashiko.dev/#/patchset/20260724-iio-adc-ti-ads112c14-burnout-v1-0-f0204cecbe3a%40baylibre.com

They have info in the cover letter that describes the prerequisites in machine-readable format.

```
---
base-commit: 0b5e142ced4bcf20532da051934bd694d1bbd470
change-id: 20260724-iio-adc-ti-ads112c14-continuous-mode-dd1580754609
prerequisite-change-id: 20260714-iio-adc-ti-ads112c14-buffered-read-41e5d1ca0dc3:v2
prerequisite-patch-id: 8723aa599dcf5b836fb81cf1efb6c9096bf60b7a
prerequisite-patch-id: 12754335ccf210bcdeaa8a5d60b2280618d50993
```

jameshilliard:
I think #389 should fix this for b4 generated patches at least.

PR #389, opened 2026-08-03
reviewer: support b4 prerequisite patches

jameshilliard:
## Summary

Apply prerequisite patches described by b4's ordered
`prerequisite-patch-id` metadata before applying and reviewing the target
series.

- Calculate and store stable Git patch IDs during patch ingestion.
- Resolve prerequisites from Sashiko's server-side database first, then fall
  back to Lore's public-inbox patch-ID search.
- Apply prerequisites in metadata order and use their resulting tip as the
  effective review baseline.
- Keep prerequisite patches out of the target review payload and preserve the
  target patch-to-commit mapping.
- Bound Lore downloads, decompressed mbox size, message count, and request
  duration.
- Migrate existing databases without an archive-wide patch-ID backfill.

## Motivation

A b4-generated series can depend on patches that have not reached an upstream
review tree yet. Sashiko currently tries to apply only the target series, so
these submissions fail even when the cover letter provides machine-readable
dependency metadata.

This addresses #49 for b4-generated patchsets containing expanded
`prerequisite-patch-id` entries. Here, "local" means Sashiko's database on the
review server, not the sender's checkout. If the server has not ingested a
prerequisite, Sashiko fetches it from Lore.

Patchew `Based-on` tags and resolving bare `prerequisite-change-id` or
`prerequisite-message-id` values without expanded patch IDs remain out of
scope.

## Validation

- `make check-all RANGE=origin/main..HEAD`
- Added coverage for metadata parsing, stable patch-ID calculation, database
  migration and lookup, duplicate ingestion, local and Lore resolution,
  response caching, exact-ID matching, and prerequisite-aware worktree
  preparation.
- Reproduced [the ACx00 failure](https://sashiko.dev/#/patchset/20260802-submit-acx00-of-dynamic-v1-v1-0-0a53cd9e21cc%40gmail.com) from its declared Linux 7.2-rc6 base: the target
  series alone fails at patch 3, while applying its four declared prerequisites
  first allows all 21 target patches to apply successfully.

## Implementation notes

- Existing patch rows retain a NULL patch ID and transparently fall back to
  Lore, avoiding an unbounded startup migration.
- Lore-only prerequisite messages are preparation inputs; they are not added
  to the ingestion pipeline or scheduled for separate reviews.
- The stored upstream baseline remains unchanged, while the transient
  prerequisite tip is passed to the review worker so range analysis excludes
  prerequisite commits.


derekbarbosa:
Hi James,

Here is some output from the /review-pr skill. I went through and removed a good deal of the comments that were hallucinations.


```
  🟡 Complexity / Logic / SOLID
   * Reviewer State Complexity: prepare_baseline_worktree in src/reviewer.rs is growing quite complex. It now handles remote fetching, worktree creation, prerequisite resolution, and patch application. 
   
   * The prerequisite-apply loop (913–947) is a distinct responsibility that should be extracted, e.g. apply_prerequisites(&worktree, &prerequisites) -> Result<String> returning the log + new head SHA, reducing the nesting in the candidate loop.
    // inside Reviewer
   +    async fn apply_prerequisites(
   +        worktree: &GitWorktree,
   +        prerequisites: &[crate::prerequisites::PrerequisitePatch],
   +    ) -> Result<(String, String)> {
   +        let mut logs = String::new();
   +        let mut head = worktree
   +            .current_head()
   +            .await
   +            .unwrap_or_else(|| "HEAD".to_string());
   +        for prerequisite in prerequisites {
   +            let mbox = git_am_message(
   +                &prerequisite.author,
   +                prerequisite.date,
   +                &prerequisite.subject,
   +                &prerequisite.diff,
   +            );
   +            worktree.apply_patch(&mbox).await.with_context(|| {
   +                format!(
   +                    "Prerequisite {} ({}) failed to apply",
   +                    prerequisite.git_patch_id, prerequisite.message_id
   +                )
   +            })?;
   +            head = get_commit_hash(&worktree.path, "HEAD").await?
   +                .with_context(|| "failed to resolve HEAD after prerequisite".to_string())?;
   +            logs.push_str(&format!(
   +                "Applied prerequisite {} ({}) as {}.\n",
   +                prerequisite.git_patch_id, prerequisite.message_id, head
   +            ));
   +        }
   +        Ok((logs, head))
   +    }
   
   * create_patch_with_git_patch_id COALESCE clears a stale id on diff change; any future caller that uses create_patch (no id) will wipe the stored id for revised patches, forcing an avoidable lore fallback. Production callers are fine today (only main.rs:2021 uses the id path), but this is a footgun given many test-only create_patch calls. Consider documenting the invariant on create_patch.

  🔵 Style / Nits / DRY
   * Duplicate Date Formatting: The PR introduces git_am_message in src/reviewer.rs to standardize git am input, which is a good cleanup of previously duplicated logic.
   * Unchecked Result in Migration: The migration in src/db.rs uses let _ = self.try_add_column(...). While try_add_column handles "already exists" errors, it's generally better to log unexpected errors rather than swallowing them entirely.
   * git_am_message uses chrono::Local for the Date: header. Same property as the old date -R subprocess it replaces (local timezone), so not a regression — but the two callers pass i64 epochs while lore-resolved prereqs carry a parsed metadata.date; the formatting branch (unwrap_or_default → empty Date:) silently emits a headerless date on Date: 0/malformed rows. Low impact.
  ---

  🔵 Style: Log migration errors instead of ignoring
  In Database::migrate, unexpected errors during column addition should be visible in logs.

   1 // src/db.rs:465
   2 -        let _ = self.try_add_column("patches", "status", "TEXT").await;
   3 -        let _ = self.try_add_column("patches", "apply_error", "TEXT").await;
   4 -        let _ = self.try_add_column("patches", "git_patch_id", "TEXT").await;
   5 +        self.try_add_column("patches", "status", "TEXT").await?;
   6 +        self.try_add_column("patches", "apply_error", "TEXT").await?;
   7 +        self.try_add_column("patches", "git_patch_id", "TEXT").await?;

  🟡 Logic: Baseline selection for prerequisites
  DESIGN_B4_PREREQUISITES.md mentions: "If a prerequisite fails on one baseline candidate, clean up that worktree and try the next candidate." The current implementation in src/reviewer.rs correctly breaks the candidate loop if prerequisites fail, but it might be worth explicitly
  logging which prerequisite failed for which baseline.

   1 // src/reviewer.rs:939
   2 +                            apply_logs.push_str(&format!(
   3 +                                "Prerequisite {} failed on baseline {}\n",
   4 +                                prerequisite.git_patch_id, baseline_sha
   5 +                            ));

```

Outside of that, my only request would be to move the 
```
#[derive(Clone)]
struct LorePatchClient {
    client: Client,
    search_url: Url,
}
```

and similar helpers for MBOX management into a separate file? maybe mbox.rs?

jameshilliard:
Changes made:

- Moved `LoreMboxClient`, bounded download and decompression, and the mbox
  splitting helpers into `src/mbox.rs`. The prerequisite resolver and existing
  API thread fetch now share the same client.
- Extracted prerequisite application into `Reviewer::apply_prerequisites`. It
  records successful prerequisites, returns the final prerequisite-tip SHA,
  and includes the patch ID and message ID in failures.
- Documented the `Database::create_patch` invariant: unchanged re-ingestion
  preserves an existing stable Git patch ID, while a changed diff clears it to
  avoid associating new content with a stale ID.

Changes not made:

- I did not replace the migration call sites with `?` because
  `try_add_column` consumes every database error and always returns `Ok(())`.
  Doing so would not surface any errors; that requires changing the shared
  migration helper and its error policy, which is outside this PR's scope.
- I did not add another baseline failure line because each attempt already
  records the baseline ref and resolved SHA, while the prerequisite error
  includes its patch ID and message ID. Both are stored in the same attempt
  log.
- I did not change date formatting because epoch `0` is valid and formats as
  January 1, 1970. Only an out-of-range timestamp produces an empty `Date:`,
  matching the previous `date -R` failure behavior.

src/mbox.rs
```diff
@@ -0,0 +1,231 @@
...
+            return Err(anyhow!("invalid prerequisite patch ID: {patch_id}"));
+        }
+
+        let mut url = self.base_url.clone();
+        url.query_pairs_mut()
+            .append_pair("x", "m")
+            .append_pair("q", &format!("patchid:{patch_id}"))
+            .append_pair("t", "1");
+        self.fetch(
+            self.client.post(url).body(""),
```
derekbarbosa, line 101: is this `post` request to search necessary? any reason why they can't be GET?

jameshilliard: Yes, POST is necessary here. Although the operation is read-only, public-inbox defines search-result mbox downloads as POST: `x=m` is explicitly ignored for GET requests ([source](https://public-inbox.org/public-inbox.git/tree/lib/PublicInbox/SearchView.pm#n217)), and its own `lei` client converts search-to-mbox URLs to POST ([source](https://public-inbox.org/public-inbox.git/tree/lib/PublicInbox/LeiInput.pm#n296)). I also verified the exact Lore endpoint: GET does not return an mbox, while POST returns `200` with `application/gzip`. The empty body is intentional because the search parameters remain in the query string.

derekbarbosa: understood. thanks for the clarification!

rgushchin:
I like the pull request overall. Few findings from gemini:
  1. Concurrency Bottlenecks: LoreMboxClient::new() builds a fresh reqwest::Client for every request. Creating a new client each time bypasses
  HTTP connection pooling, adding significant latency and connection overhead under load.
  2. Async Deadlock Risk (git patch-id): The logic attempts to asynchronously write the entire patch diff to stdin while awaiting completion.
  If a massive diff is passed and the child process outputs enough data to fill the OS pipe buffer (~64KB) before reading all of stdin, both
  processes will block indefinitely (a known Tokio async anti-pattern).
  3. Timestamp Decoding Flaw: Error handling during date parsing falls back to an empty string (""). Passing an empty date header to git am
  will silently corrupt baseline preparation and trigger confusing failures.
  4. Missing Transient Retry Logic: If Lore responds with an HTTP 503 or a timeout, the system immediately marks the baseline as Failed. It
  lacks graceful retry scaffolding, which will generate noise for developers if Lore experiences momentary hiccups.

jameshilliard:
Implemented findings 1, 3, and 4 in the amended implementation commit `77715a1`:

- `LoreMboxClient` now reuses a process-wide `reqwest::Client`, preserving connection pooling across Lore requests.
- Lore requests now make two bounded retries, after one and two seconds, for connection, timeout, request, and response-body failures as well as HTTP 408, 429, and 5xx responses. Permanent failures are returned immediately. Tests cover recovery from a transient 503, no retry for a 404, and exhaustion after three total attempts.
- Mbox generation now returns an error for timestamps outside Chrono's supported range instead of producing an empty `Date:` header. Tests cover both the valid Unix epoch and an out-of-range timestamp.

I did not change finding 2. Each `git patch-id` process receives one patch and produces only one small result line, so stdout cannot fill its pipe while the patch is being written. The stdin handle is dropped after `write_all`, which signals EOF, and `wait_with_output` then drains stdout and stderr. Concurrent stdin/stdout handling is necessary for subprocess protocols that can produce unbounded output before consuming their input, but that condition is not present here.

jameshilliard:
rebased

derekbarbosa:
@rgushchin hey roman, pinging for a manual merge.

jameshilliard:
rebased

jameshilliard:
rebased again

Issue #274, opened 2026-06-12
Cannot use OpenRouter with openai-compatible provider


xzz53:
Hi!

I'm trying to set up sashiko in local mode using OpenRouter as an LLM provider. Corresponding part of config I use looks like

```toml
[ai]
# Can be "openai" or "openai-compatible"
provider = "openai-compatible"
model = "anthropic/claude-opus-4.6

[ai.openai_compat]
base_url = "https://openrouter.ai/api/v1/chat/completions"
context_window_size = 8192
max_tokens = 2048
```

I run it like this: `sashiko-cli local --force-local HEAD` and get the following output:

```
[1/4] Extracting patches from HEAD... (1 commit)
[2/4] Starting review subprocess...
[2/4] Applying patches to worktree...
[3/4] AI review in progress...
Local Review Results
  Input:    HEAD
  Baseline: 3c056cb668b235d3768c6e41dc5897823455cd97^

Patch Application:
  [1] applied (checkout)

Error: Stage 1 failed to produce valid 'concerns' and 'dismissed_concerns' arrays after 3 attempts — aborting review
```

After studying the sources, I found that this line in openai::normalize_base_url() may be the culprit:

https://github.com/sashiko-dev/sashiko/blob/8d35e352c576f6ef27e5e151a956260e65898d0f/src/ai/openai.rs#L209

TheSven73:
I ran into this as well, and the fix is pretty simple:

```diff
diff --git a/src/ai/openai.rs b/src/ai/openai.rs
index 637f386..6c27ee2 100644
--- a/src/ai/openai.rs
+++ b/src/ai/openai.rs
@@ -206,6 +206,7 @@ impl OpenAiCompatClient {
         let path = match path.as_str() {
             "" => "/chat/completions",
             "/v1" | "/v1/chat/completions" => "/v1/chat/completions",
+            "/api/v1" | "/api/v1/chat/completions" => "/api/v1/chat/completions",
             _ => return String::new(),
         };
```

xzz53:
Sure, fixing this specific case is trivial. But honestly, I can't understand the motivation behind commit d5b6fd1, which introduced this behavior. If we'd like all URLs to end in "/chat/completions", maybe should check that at first, instead of breaking things?..

> I ran into this as well, and the fix is pretty simple:
> 
> diff --git a/src/ai/openai.rs b/src/ai/openai.rs
> index 637f386..6c27ee2 100644
> --- a/src/ai/openai.rs
[...]

PR #391, opened 2026-08-04
ai: accept full chat completions URL in openai-compat base_url

SquisherDerHeld:
## Problem

The `openai-compatible` provider's `normalize_base_url()` accepts only a fixed set of path shorthands (bare host, `/v1`, `/api/v1`) and rejects every other path with `Invalid OpenAI url`:

```rust
let path = match path.as_str() {
    "" => "/chat/completions",
    "/v1" | "/v1/chat/completions" => "/v1/chat/completions",
    "/api/v1" | "/api/v1/chat/completions" => "/api/v1/chat/completions",
    _ => return Err(anyhow::anyhow!("Invalid OpenAI url {}", url)),
};
```

This blocks providers whose endpoint path is not on that allowlist. The most notable case is **z.ai / Zhipu** (`glm-*` models), whose OpenAI-compatible gateways use:

- `…/api/paas/v4/chat/completions` — direct API (billed to the API resource package)
- `…/api/coding/paas/v4/chat/completions` — coding-plan gateway (billed to the coding-plan subscription)

Both are rejected today. `default_base_url_for_model()` already special-cases `glm-*` to the China endpoint `open.bigmodel.cn`, but the international `api.z.ai` endpoint and its separately-billed coding-plan gateway cannot be configured at all.

## Solution

Instead of extending the per-provider allowlist, accept any `base_url` that already targets a chat completions endpoint verbatim:

```rust
// If the caller supplied a full URL that already targets a chat
// completions endpoint, accept it verbatim.
if path.ends_with("/chat/completions") {
    return Ok(format!("{base}{path}"));
}
```

This makes **any** OpenAI-compatible provider configurable via `base_url` alone, while the existing shorthand matching (and rejection of unknown shorthand paths) is fully preserved. It's strictly more permissive for explicit full URLs and changes no existing behaviour.

z.ai then works purely from config:

```toml
[ai]
provider = "openai-compatible"
model = "glm-5.2"

[ai.openai_compat]
base_url = "https://api.z.ai/api/coding/paas/v4/chat/completions"
context_window_size = 128000
max_tokens = 16384
```

## Changes

- `src/ai/openai.rs`: passthrough for full `/chat/completions` URLs + unit tests.
- `docs/llm-providers.md`: document both `base_url` forms and add a z.ai coding-plan example.

## Notes

- `glm-5.2` is a reasoning model whose thinking tokens count against `max_tokens`; the doc example uses 16384 to avoid truncation during review.

## Testing

- `cargo test --lib` (new + existing `normalize_base_url` tests pass)
- `cargo clippy --all-targets` clean
- `cargo fmt --check` clean

### DIFF
--- a/docs/llm-providers.md
+++ b/docs/llm-providers.md
@@ -279,3 +279,30 @@ cp docs/examples/Settings.openai-compat.toml Settings.toml
 ```
 
 Adjust `base_url` to point to your provider's endpoint.
+
+`base_url` may be either:
+
+- a shorthand such as `http://localhost:8080/v1` (the `/chat/completions`
+  suffix is appended automatically), or
+- the full chat completions URL, e.g.
+  `https://api.z.ai/api/coding/paas/v4/chat/completions`. Use this form for
+  providers whose path is not a recognised shorthand.
+
+**z.ai / Zhipu (glm-*) example:**
+
+z.ai exposes two OpenAI-compatible gateways with **separate billing**: a
+direct API (`…/api/paas/v4/…`, billed to the API resource package) and a
+coding-plan gateway (`…/api/coding/paas/v4/…`, billed to the coding-plan
+subscription). To review against the coding-plan quota, point `base_url` at
+the full coding endpoint and set `LLM_API_KEY` (or `OPENAI_API_KEY`):
+
+```toml
+[ai]
+provider = "openai-compatible"
+model = "glm-5.2"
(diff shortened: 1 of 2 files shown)


derekbarbosa:
pinging @Rubusch, @TheSven73 and @xzz53 for a hand in review here from their recent involvement in #274  #275  and #308

PR #393, opened 2026-08-05
reviewer: surface git am error in baseline apply log

### LINKED ISSUE #388: Provide more feedback when patchset "Failed To Apply"
When a patch set fails to apply, the Baseline "details" has a "view log" which given very few extra details.
I would like to see the error from "patch" or similar.  Because...

https://sashiko.dev/#/patchset/20260803012648.2195405-1-neilb%40ownmail.net

claims that it tries against the exact base that I used 
  075b74841bd0065a3bda3440873c747938e69b68
and I would like to know why that failed.

derekbarbosa:
The baseline apply loop dropped the git am error via is_ok(), so the
View Log only showed a generic failure line with no reason. Capture the
error from apply_patch and append it to the baseline log so the actual
git am stdout/stderr is visible when a patch fails to apply.

Fixes: #388 

Assisted-by: deepseek-v4-flash

### DIFF
--- a/src/reviewer.rs
+++ b/src/reviewer.rs
@@ -928,8 +928,13 @@ impl Reviewer {
                     );
 
                     // Try git am
-                    if (worktree.apply_patch(&mbox).await).is_ok() {
-                        applied = true;
+                    match worktree.apply_patch(&mbox).await {
+                        Ok(_) => applied = true,
+                        Err(e) => {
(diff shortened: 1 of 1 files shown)


rgushchin:
No, this is not the direction I want to go. It opens a path to various security vulnerabilities. Sashiko logs the baseline commit, so it's easy to reproduce the problem locally.

PR #397, opened 2026-08-06
fix: better distinguish pre-existing findings 

### LINKED ISSUE #381: Improve propagation of pre-existing issues
Related Thread: https://lore.kernel.org/all/CAJnrk1Y6twWcXfFesbLMNfMavg5RWgQ07pGJ1BM98V4o+FYcBw@mail.gmail.com/

> Currently, they're flagged on patch reviews, which can be
> a bit frustrating for contributors who didn't introduce the problem,
> and adds a bit of noise for maintainers/reviewers looking at the
> Sashiko report.

Thinking of a way to separate pre-existing issues found during report generation to a separate section...

derekbarbosa:
Modify the Stage 11 prompt and SPA to produce better visual clarity on pre-existing findings.

Fixes: #381

### DIFF
--- a/src/worker/prompts.rs
+++ b/src/worker/prompts.rs
@@ -313,11 +313,15 @@ You are the lead reviewer validating consolidated concerns. You will be given a 
 
 You are an automated review bot generating a report for the Linux Kernel Mailing List (LKML). Convert the provided JSON findings into a polite, standard, inline-commented LKML email reply.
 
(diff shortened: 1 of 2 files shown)


rgushchin:
First, we don't need to spend tokens on this, earlier stages determine if an issue is pre-existing, so we can simple hide them from later stages. However I'm not sure it's the right thing to do.
Once we'll have a database of pre-existing issues and it will be actively used by maintainers, we'll stop reporting pre-existing issues together with new issues. As of now, let's keep it as it is.

derekbarbosa:
thanks, Roman

Issue #249, opened 2026-06-05
Feature Request: Enrich Stage 10 cross-patch context to reduce false positives


m0ck1ng:
## Problem

Sashiko's review pipeline reviews each patch in isolation. **Stages 1–7 have zero visibility into sibling patches.** Stage 10 is the only stage with cross-patch awareness. It has access to git tools (read files, diff, show, grep, etc.) and could theoretically inspect subsequent patches' code. However, the only cross-patch information *injected into its prompt* is the **subject lines** of subsequent patches.

This creates a systematic class of false positives: findings on patch N that are actually resolved or implemented by a later patch M in the same series. Three common patterns:

1. **Stub + Implementation split**: Patch 1 introduces a function stub; patch 3 implements it. Review of patch 1 reports "missing implementation" — which is intentional, not a bug.

2. **Transient bug + Fix**: Patch 3 introduces a race condition during refactoring; patch 5 fixes it. Review of patch 3 reports the bug — which doesn't exist in the patchset's final state. This is noise for maintainers.

3. **Incremental construction**: A series builds up a feature step by step (parse → validate → integrate). Each intermediate patch appears "incomplete" because the next patch hasn't been applied yet. Isolated review treats every incremental step as buggy.

Stage 10 is designed to catch these cases — it has a validation step that checks whether concerns are addressed by subsequent patches, and it has git tools to inspect the code. But with only subject lines as starting cues, its catch rate is effectively zero (1.6%, see data below). The issue is not that Stage 10 lacks tools; it's that subject lines don't provide enough signal for the LLM to know *what to look up* with those tools.

## Data

Analysis of recent 1,271 reviewed patchsets (≥4 patches, with findings) from the sashiko.dev API:

| Metric | Value |
|--------|-------|
| Total findings | 13,790 |
| Cross-patch FP candidates | 3,534 (**25.6%** of all findings) |
| Patchsets affected | 195 (**15.3%** of analyzed patchsets) |
| Non-preexisting FPs | 2,386 (67.5% of candidates — AI flagged them as newly introduced) |
| Stage 10 cross-patch catch rate | 57 / 3,534 (**1.6%**) |

**Severity distribution**: 79.6% of cross-patch FP candidates are High or Critical — the findings that waste the most maintainer time.

**Patch distance**: 54.7% of FPs are within ±2 patches; 24.7% are in adjacent patches.

## Examples

[...]

PR #400, opened 2026-08-07
RFC: Introduce SeriesMap

### LINKED ISSUE #392: Improve cross-patch issue detection
From https://lore.kernel.org/sashiko/ed23ce9e-2cdd-472a-85e4-c0cdf81359b9@linux.ibm.com/T/#t

> My concern is more that we get reports about issues that are fixed in a subsequent patch.

derekbarbosa:
Hi, 

I was debating sending this RFC up, but given the recent issues, I decided to rebase and implement some additional fixups.

---

Parallel workers review each patch in isolation. Patch N would
flag an incomplete interface that Patch N+2 completed, or an intermediate
breakage that a later fixup corrected.

Introduce SeriesMap in a two stage pipeline.

Before spawning workers, a pre-computation stage generates a SeriesMap and a human-readable
summary. Both are injected into every worker's prompt with suppression rule.

Single-patch series skip the extra calls. On SeriesMap failure, workers get a lightweight sibling-patch subject list as fallback.

Fixes: #392 #249


rgushchin:
For each finding in each patch in the series which is not the last patch Sashiko verifies (or at least should verify) if the finding is resolved by the following changes in the series. It should be done based on the code, not some short summary of the patchset, otherwise we'll see a lot of false positives. So I'm not sure we need something like SeriesMap tbh.

PR #401, opened 2026-08-08
test: make bad-tag recovery test robust to git version

helq:
git_ops::tests::test_ensure_remote_bad_tag_recovery asserted that a corrupt loose tag ref is physically deleted by ensure_remote's recovery path. That path only runs when `git fetch` fails with "fatal: bad object refs/tags/...", which older git emits but current git (>= 2.x, verified on 2.55) does not: modern git silently ignores the malformed loose ref and the fetch succeeds, so the tag file lingers and the assertion fails. The test was green in upstream CI only because CI ran an older, stricter git.

Assert the version-independent guarantee instead: after ensure_remote, the fetch has succeeded (remote-tracking refs exist) and the bad tag is not a live, valid ref. This holds on both old git (recovery deletes the tag) and new git (the malformed ref is ignored and the fetch succeeds).

### DIFF
--- a/src/git_ops.rs
+++ b/src/git_ops.rs
@@ -1450,7 +1450,13 @@ mod tests {
         let mut bad_tag_file = File::create(&bad_tag_path)?;
         writeln!(bad_tag_file, "0000000000000000000000000000000000000000")?;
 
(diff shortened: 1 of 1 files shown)


rgushchin:
Thanks!

PR #409, opened 2026-08-10
feat: prefer local Settings.toml over embedded defaults

liuhangbin:
In the project README.md, it said

"""
  To get started with the default (Gemini):

  cp docs/examples/Settings.example.toml Settings.toml

  [ ... ]

  sashiko init
"""

Which gives users an impression that the init will use Settings.toml. But it actually uses the Settings.example.toml by default.

When running 'sashiko init', let's check for a Settings.toml in the current directory first and use its contents if found. Falls back to the embedded DEFAULT_SETTINGS otherwise.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -901,7 +901,15 @@ fn handle_init_command(
         std::fs::create_dir_all(parent)?;
     }
 
-    std::fs::write(&path, DEFAULT_SETTINGS)?;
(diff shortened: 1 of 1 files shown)


rgushchin:
No, I think it's not a good idea. Instead the documentation should be fixed. sashiko init is created for a case when there is no Settings.toml because of the cargo-driven installation, otherwise Sashiko will already prefer a local settings file.

PR #411, opened 2026-08-10
API deduplicate api/submit requests

helq:
Harden the API ingestion pipeline against duplicate and redundant submissions that arise when multiple api/submit.

This occurs when we submit the same SHA for review over and over from our CI/CD to sashiko

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use crate::db::Database;
-use crate::events::Event;
+use crate::events::{Event, MessageSource};
 use crate::fetcher::FetchRequest;
 use axum::{
     Json, Router,
@@ -358,7 +358,7 @@ fn generate_synthetic_id(prefix: &str) -> String {
         .expect("Time went backwards");
     // e.g. sashiko-local-1715890000-12345
     format!(
-        "sashiko-{}-{}-{}",
+        "sashiko-{}-{}-{}@sashiko.local",
         prefix,
         since_the_epoch.as_secs(),
         fastrand::u32(..)
@@ -399,6 +399,8 @@ async fn submit_patch(
 
             let event = Event::RawMboxSubmitted {
                 raw,
+                submission_id: id.clone(),
+                source: MessageSource::ApiInject,
                 group: "api-submit".to_string(),
                 baseline: base_commit,
                 skip_subjects,
@@ -438,7 +440,7 @@ async fn submit_patch(
             if let Err(e) = state
                 .db
                 .create_fetching_patchset(
-                    &id,
+                    &format!("{}@sashiko.local", id),
                     &format!("Fetching {} from {}...", &sha, repo_display),
                     skip_subjects.as_ref(),
                     only_subjects.as_ref(),
@@ -523,6 +525,7 @@ async fn submit_patch(
                         .send(Event::IngestionFailed {
                             article_id: msgid_clone.clone(),
                             error: format!("Failed to fetch thread: {}", e),
+                            source: MessageSource::ApiFetchThread,
                         })
                         .await;
                 }
@@ -594,6 +597,8 @@ async fn fetch_and_inject_thread(
 
     let event = Event::RawMboxSubmitted {
         raw,
+        submission_id: msgid.to_string(),
+        source: MessageSource::ApiFetchThread,
         group: "api-submit".to_string(),
         baseline: None,
         skip_subjects: None,
@@ -1234,4 +1239,16 @@ async fn forge_webhook(
         "status": "accepted",
         "message": format!("{} {} queued for review", forge.name(), action)
     })))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_generate_synthetic_id_format() {
+        let id = generate_synthetic_id("test");
+        assert!(id.starts_with("sashiko-test-"));
+        assert!(id.ends_with("@sashiko.local"));
+    }
 }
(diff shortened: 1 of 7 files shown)


rgushchin:
🤖 **AI Code Review / Regression & Security Analysis**

Here is an automated review of PR #411 (`Upstream/api dedup`):

---

### Summary of Changes
PR #411 aims to harden the patch ingestion pipeline against duplicate and redundant submissions (e.g. repeated CI/CD triggers for identical commit SHAs or ranges).

---

### 🚨 Critical Regression Risks & Functional Flaws

#### 1. Missing Physical Patch Rows in Cross-Patchset / Range Resubmissions (Critical)
* **Location:** `src/db.rs` (`create_patch`)
* **Mechanism:**
  When a commit SHA that was already ingested in another patchset (e.g. first as a singleton or in another branch) is submitted as part of a new patchset, `Database::create_patch` skips inserting a row into the `patches` table for the new patchset to avoid moving/stealing the row from the previous patchset. Instead, it only increments `received_parts` (a "virtual" count).
* **Impact:**
  1. The new patchset has `received_parts == total_parts` and transitions to `Pending`, but does **not** physically own the patch row in the database.
  2. When `Database::get_patch_diffs(patchset_id)` runs (`SELECT ... FROM patches WHERE patchset_id = ?`), it only returns physical patch rows for that `patchset_id`.
  3. In `Reviewer::review_patchset_task`, the reviewer attempts to apply the series to the baseline git tree. Because the shared commit is missing from the list:
     - Subsequent dependent patches in the series will fail git application, causing the entire patchset to fail with `FailedToApply`.
     - Or, if it applies, the shared patch is completely omitted from the review, leaving the patchset only partially reviewed.
  4. Both `Database::get_patchset_details` (UI & API) and `Database::get_completed_reviews_for_release` (notifications/outbox) query `FROM patches WHERE patchset_id = ?` and will omit the patch from the new patchset.
* **Root Cause:**
  In `src/schema.sql`, the `patches` table has `message_id TEXT NOT NULL UNIQUE`. Because Git commit SHAs are reused across different branches/submissions as `message_id`, the database constraint forbids the same SHA from existing as a patch in multiple patchsets. A "virtual" counter bypasses the constraint but breaks all subsequent relational queries.
* **Suggested Fix:**
  Update the database schema constraint so patches are scoped per patchset (e.g., `UNIQUE(patchset_id, message_id)` or `UNIQUE(patchset_id, part_index)`), or namespace `message_id` for Git commits (e.g., `ps_{patchset_id}_{sha}`), ensuring each patchset physically owns its patch records.

---

#### 2. Synthetic ID Suffix (`@sashiko.local`) Mismatch in Duplicate Check (High)
* **Location:** `src/api.rs` (`submit_patch`) & `src/db.rs` (`has_patchset_by_msgid`)
* **Mechanism:**
  In `src/api.rs`, `submit_patch` calls `state.db.has_patchset_by_msgid(&id)` where `id` is the bare SHA or range string.
  However, placeholders created by `create_fetching_patchset` and patchsets ingested via `git-fetch` have `cover_letter_message_id = format!("{}@sashiko.local", id)`.
* **Impact:**
  - `Database::has_patchset_by_msgid` checks `cover_letter_message_id = ? OR cover_letter_message_id = <?>` without checking the `@sashiko.local` candidate. As a result, the optimistic duplicate check **never matches** remote fetch patchsets.
  - Similarly, `Database::get_patchset_details_by_msgid` does not check `@sashiko.local`, causing `GET /api/patch?id=<sha>` to return `404 Not Found` while the remote patchset is in `Fetching` status.
* **Suggested Fix:**
  Update `has_patchset_by_msgid` and `get_patchset_details_by_msgid` to query candidate IDs including `format!("{}@sashiko.local", msgid)` and `msgid.strip_suffix("@sashiko.local")`.

---

#### 3. Indiscriminate Short-Circuiting in Duplicate Remote Check (Medium)
* **Location:** `src/api.rs` (`submit_patch`)
* **Mechanism:**
  `Database::has_patchset_by_msgid` checks `SELECT 1 FROM patchsets WHERE cover_letter_message_id = ?` regardless of `status`.
* **Impact:**
  If a previous fetch had `status = 'Failed'` or `'Cancelled'`, an optimistic duplicate check that matches will immediately return `accepted` and skip queuing the fetch. This breaks the retry capability introduced in `create_fetching_patchset` that resets `Failed` and `Cancelled` records to `Fetching`.
* **Suggested Fix:**
  Only skip the fetch in `submit_patch` if the existing patchset is in an active or completed status (`Pending`, `In Review`, `Reviewed`, `Incomplete`), allowing `Failed` and `Cancelled` patchsets to fall through to `create_fetching_patchset` for retry.

---

### 🔒 Security Analysis

* **SQL Injection:**
  ✅ Safe. All new and modified SQL statements use parameterized queries via `libsql::params!`.
* **Race Conditions:**
  ✅ Synchronous status update to `In Review` in `Reviewer::process_pending_patchsets` before spawning background review tasks correctly prevents tight reviewer polling loops from double-dispatching the same patchset.
* **Author Validation:**
  ℹ️ `is_strict_author` disables strict author verification for series submitted via `ApiInject` and `ApiFetchThread` (`total_parts > 1`). While necessary to support legitimate multi-author kernel series, ensure that `/api/submit` endpoints are protected by appropriate access controls if deployed in multi-tenant environments.
* **Randomness:**
  ℹ️ `generate_synthetic_id` uses `fastrand` (PRNG). This is adequate for internal synthetic ID collision avoidance, provided these IDs are not expected to be unguessable security tokens.

helq:
These are indeed very good points. I have seen issue 1 pop up internally a couple of times, but it never was an issue. After some thought and with Gemini reminding me that reviews of the same patch across different patchsets have different contexts, I was convinced to fix all issues.

rgushchin:
Rebased and merged to main in commit bb6eac1da848.

PR #412, opened 2026-08-10
docker: prevent deletion of queued commits on container restart

helq:
Removing the aggressive git reflog expire and git prune commands from the container startup script.

Previously, these commands immediately deleted all unreachable commits from the persistent git repository. Because pending reviews in the database queue are not pointed to by any git branch or tag, they were marked as unreachable and pruned during restarts, leading to Failed to Apply errors when the workers attempted to start reviews.

Relying on git gc --auto instead, which safely preserves unreachable commits for the default 2-week grace period.

### DIFF
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -49,9 +49,7 @@ setup_kernel() {
         # Prune unreachable objects and run gc to keep the repo healthy
         echo "Pruning and garbage collecting..."
         git remote prune origin || true
-        git reflog expire --all --expire=now
         git gc --auto
-        git prune
         # Ensure we are on master and it's clean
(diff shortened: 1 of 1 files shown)


helq:
Yes, very clearly as mentioned before by roman in some other thread: adding stuff to the docker startup file instead of sashiko-bin has made this part of the code very fragile. Instead of adding new functionality I am removing it because it has broken the service on restartup (if there are pending reviews for SHAs that were fetched from a remote already).

rgushchin:
I don't see much harm and going to merge it, however I doubt it's a reliable design overall, I believe we need to create temp branches or tags in such cases to make 100% sure we retain useful git objects. Something to think about.

PR #422, opened 2026-08-11
prompts: handle Runtime PM conditional gets

### LINKED ISSUE #396: False positive around Runtime PM get()/put()
Sashiko expects `pm_runtime_put()` to be called if `pm_runtime_get_if_in_use()` or `pm_runtime_get_if_active()` return a negative error value. Search for `pm_runtime_put()` in any recently posted camera sensor driver:
https://sashiko.dev/#/patchset/20260520-imx678-v3-0-8b5f9676486e%40ideasonboard.com
https://sashiko.dev/#/patchset/20260730021525.166811-1-lachlan.michael%40sony.com

Sashiko's suggestion is invalid, as `pm_runtime_put()` is a no-op when CONFIG_PM=n, see:
https://github.com/torvalds/linux/blob/master/include/linux/pm_runtime.h#L261-L263

Some developers might read that and also infer that they should return early if `pm_runtime_get_if_active()` returns a negative error code, like done in v2 of the imx908 series, which leads to another Sashiko complaint:
https://sashiko.dev/#/patchset/20260806070934.21764-1-lachlan.michael%40sony.com

It's quite annoying that it complains in both cases, without a suggested "fix". And the suggested "fix" is actually to keep things as they were originally:
```c
if (!pm_runtime_get_if_active(&client->dev))
    return 0;

/* write registers */

/* unconditional put */
pm_runtime_put(&client->dev)
```

OllieinCanada:
Fixes #396

## Summary

Teach Sashiko's power-management prompt to account for the configuration-dependent return contract of `pm_runtime_get_if_active()` and `pm_runtime_get_if_in_use()`.

## Root cause

With `CONFIG_PM=n`, these helpers return a negative stub value while Runtime PM put operations are no-ops. Treating every negative return as an error that requires an early return can incorrectly suppress hardware operations in kernels built without Runtime PM.

## Changes

- distinguish positive, zero, and negative conditional-get results
- require proof from the exact helper, configuration, and device lifecycle
- preserve reference-balance warnings when Runtime PM may be disabled for a device under `CONFIG_PM=y`
- avoid applying this exception to helpers with different contracts
- correct the existing IRQ example so it does not recommend `ret <= 0` unconditionally

## Validation

- verified the behavior against the upstream Linux Runtime PM header and implementation
- `git diff --check`
- signed-off commit verified
- GitHub Actions DCO, lint, and Rust unit-test jobs pass

### DIFF
--- a/third_party/prompts/kernel/subsystem/pm.md
+++ b/third_party/prompts/kernel/subsystem/pm.md
@@ -38,6 +38,56 @@ and `pm_runtime_get()`.
 - Returns **0** on successful resume or when an async request was queued
 - Returns negative errno on failure
 
+## Conditional Gets When `CONFIG_PM=n`
+
+Do not treat a negative return from `pm_runtime_get_if_active()` or
+`pm_runtime_get_if_in_use()` as proof that the caller must return early or
+call `pm_runtime_put()`. These helpers have configuration-dependent contracts:
+
+- A positive return means the helper incremented the usage counter. The caller
+  must pair that reference with a put on every subsequent exit path.
+- Zero means the condition was not met and the counter was not incremented.
+  The caller normally skips hardware access and must not put a reference it did
+  not acquire.
+- A negative return also means the counter was not incremented. With
+  `CONFIG_PM=y`, `-EINVAL` reports that Runtime PM is disabled for the device.
+  With `CONFIG_PM=n`, the inline stubs return `-EINVAL`, while
+  `pm_runtime_put()` and the other counter helpers are no-ops.
+
+This makes the following driver pattern potentially intentional:
+
+```c
+int ret = pm_runtime_get_if_active(dev);
+
+if (!ret)
+    return 0;
+
+/* Access hardware. */
+
+pm_runtime_put(dev);
+```
+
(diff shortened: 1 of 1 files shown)


rgushchin:
LGTM, but it's in a draft state, should I move on and merge it? Or there is something to wait for?

OllieinCanada:
Thanks! Nothing else is pending from my side. I have marked the PR ready for review; please feel free to merge it.

rgushchin:
Cool, thanks!
Can you, please, fix the SOB line with the real name and email?
Thanks

OllieinCanada:
Updated the commit with my real author and Signed-off-by identity. DCO, lint, and unit tests are all green, and nothing else is pending from my side.

derekbarbosa:
@rgushchin gentle ping here.

OllieinCanada:
Rechecked the current head after the earlier sign-off repair: the single commit is authored by Oliver Slapinski <olliefromcanada@gmail.com>, has the matching Signed-off-by line, and all DCO, lint, and unit-test checks are green. Nothing remains pending from my side.

PR #425, opened 2026-08-12
prompts: verify cross-function resource cleanup

### LINKED ISSUE #124: Bug: False Positive in Resource Cleanup Detection
# Observed Behavior
The tool flagged a missing resource cleanup that was actually present, likely due to a lack of deep call-graph analysis.

## Technical Requirement
Improve Sashiko's ability to track resource state across function boundaries. If a cleanup function is called within a conditional block, the tool could verify the contents of that function before flagging a leak.

OllieinCanada:
Fixes #124

## Summary

Require Sashiko to inspect cleanup helper implementations before reporting a cross-function resource leak.

## Root cause

The resource-management prompts required ownership tracking across function boundaries, but the final false-positive verification did not explicitly require opening cleanup, unwind, destroy, unregister, or release helpers. A call site could therefore appear to leak a resource even when a conditional helper released it internally.

## Changes

- inspect cleanup helper bodies and relevant callees rather than inferring behavior from names
- evaluate the condition guarding a cleanup call on the exact candidate path
- track the same object, field, reference, descriptor, or registration from acquisition to release
- account for nested helpers, managed resources, ownership transfer, and subsystem unregister behavior
- retain real findings when cleanup is bypassed, partial, applies to another resource, or cannot be verified
- require an auditable acquisition-to-release verdict

## Validation

- confirmed the guide is always loaded during Stage 10 false-positive verification
- `git diff --check`
- signed-off commit verified

No model-backed regression or paid API call was used.

### DIFF
--- a/third_party/prompts/kernel/false-positive-guide.md
+++ b/third_party/prompts/kernel/false-positive-guide.md
@@ -166,6 +166,35 @@ the code behaves a certain way, you MUST verify against the actual implementatio
 - Understand subsystem ownership models
   - Output: quote subsystem convention or "no documented model"
 
+### 6.1. Cross-Function Cleanup Verification (MANDATORY)
+
+A resource is not leaked merely because the current function does not release
+it directly. Before retaining a leak concern, inspect every cleanup, unwind,
+destroy, unregister, or release helper reached by the candidate exit path.
+Open the helper body and follow its relevant callees far enough to determine
+what happens to the exact resource. Do not infer behavior from the helper's
+name alone.
+
+For a cleanup helper called conditionally, prove whether the condition is true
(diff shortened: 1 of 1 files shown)


rgushchin:
This look reasonable to me, but do you have any examples or stats on how it helps in practice? Any specific false positives it helps to avoid? I strongly prefer to make prompts changes based on data.

OllieinCanada:
I don't have aggregate stats. The source is #124, but that report doesn't include the affected patchset, code, or exact false-positive output. I agree that isn't enough evidence for a confident prompt change, so I've moved this back to draft and asked for a concrete reproducer before taking it further.

PR #430, opened 2026-08-13
toolbox: fix panic reading a file that is empty at the revision

ftabba:
`git_read_files` clamps `start_line` and `end_line` to `[1, total_lines]` before working out the slice bounds. When the file is empty at the requested revision `total_lines` is 0, so the clamp's lower bound is above its upper bound and it panics:

    assertion failed: min <= max, core/src/cmp.rs:1082

This raises the clamp's upper bound to `total_lines.max(1)`, so its lower bound (`1`) can never exceed its upper bound and the clamp cannot panic. Nothing changes for a non-empty file, where `total_lines.max(1)` equals `total_lines`. For an empty file the code after the clamp already handles it: the `match` caps `end` at `total_lines`, the next two lines bring `start` and `end` into range, and the `start >= total_lines` guard returns an empty result instead of slicing.

Removing the two pre-clamp lines would also avoid the panic, but it changes what an out-of-range `start_line` returns (the last line today, an empty result without them), so I kept the smaller change.

The second commit adds a regression test: it reads a file that is empty at the revision with a line range set and checks the tool returns empty content rather than panicking. Without the fix the test aborts on the same assertion.

### DIFF
--- a/src/toolbox/git_read_files.rs
+++ b/src/toolbox/git_read_files.rs
@@ -163,8 +163,11 @@ impl GitReadFilesTool {
         let lines: Vec<&str> = content.lines().collect();
         let total_lines = lines.len();
 
-        let start_line = start_line.map(|s| s.clamp(1, total_lines));
-        let end_line = end_line.map(|e| e.clamp(1, total_lines));
+        // total_lines is 0 for an empty file, and clamp() panics when its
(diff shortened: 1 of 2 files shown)


rgushchin:
Merging, thanks!

PR #431, opened 2026-08-14
README: "a set *of* Linux..." instead of "a set Linux..."

revdefeatist:
A simple grammatical correction.


rgushchin:
Can you, please, add your SOB (Signed-off-by) line?
Thanks

revdefeatist:
There you go. If there are any issues, you can feel free to just make the correction yourself. I don't normally use github but was looking at this repo and caught a little typo and corrected it in-browser, to see what that feature is like. I don't really care about attribution, and I might not login here for a while.

PR #435, opened 2026-08-15
db: make outbox inserts atomic

OllieinCanada:
## Summary

Make Sashiko's notification outboxes idempotent under concurrent review completion.

- replace separate deduplication lookups and inserts with one conditional insert per outbox path
- cover review email, Patchwork status update, and Patchwork failure notification queues
- add deterministic tests that release 32 concurrent writers for each logical notification and verify exactly one row is stored

## Root cause

Each outbox writer first queried for an existing row and then issued a separate insert. Two review tasks completing together could both observe that no row existed before either insert ran, allowing duplicate external notifications to be queued.

The code described these paths as duplicate-safe, but the invariant was not enforced atomically at the persistence boundary.

## Design

Each writer now uses a single `INSERT ... SELECT ... WHERE NOT EXISTS` statement. SQLite/libSQL serializes the conditional check and write as one statement, closing the check-then-act window without a schema migration or destructive cleanup of historical outbox rows.

The existing logical deduplication keys remain unchanged:

- review email: patch ID
- Patchwork update: message ID, API URL, and context
- Patchwork notification email: recipient, subject, and reply target

## Validation

- standalone `rustfmt --check` passes for `src/db.rs`
- `git diff --check` passes
- direct SQLite concurrency probe passes for all three statements with 32 separate writers
- Rust concurrency tests are included for all three outbox paths
- local `cargo test` and `make check-pr` stop before compiling Sashiko because `aws-lc-sys` rejects the installed GCC 9.4 compiler; GitHub CI is expected to provide the authoritative Rust, clippy, and test results

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -3881,39 +3881,36 @@ impl Database {
         references_hdr: &str,
         body: &str,
     ) -> Result<()> {
-        // Prevent duplicate emails for the same patch
-        let mut rows = self
+        let created_at = chrono::Utc::now().timestamp();
+        let inserted = self
             .conn
-            .query(
-                "SELECT 1 FROM email_outbox WHERE patch_id = ?",
-                libsql::params![patch_id],
+            .execute(
+                "INSERT INTO email_outbox (patch_id, status, to_addresses, cc_addresses, subject, in_reply_to, references_hdr, body, created_at)
+                 SELECT ?, ?, ?, ?, ?, ?, ?, ?, ?
+                 WHERE NOT EXISTS (
+                     SELECT 1 FROM email_outbox WHERE patch_id = ?
+                 )",
+                libsql::params![
+                    patch_id,
+                    status,
+                    to_addresses,
+                    cc_addresses,
+                    subject,
+                    in_reply_to,
+                    references_hdr,
+                    body,
+                    created_at,
+                    patch_id,
+                ],
             )
             .await?;
 
-        if let Ok(Some(_)) = rows.next().await {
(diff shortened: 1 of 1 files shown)


OllieinCanada:
Resolved the conflict by merging current upstream main in signed commit a04f40d. The conflict was limited to the src/db.rs test-module tail; the resolution retains both upstream's new database tests and all three concurrent outbox regressions. The cumulative PR diff remains src/db.rs only, cargo fmt --check and git diff --check pass, and exact-head Linux CI is now authoritative for the full Rust suite.

derekbarbosa:
Hey Ollie, could you run some tests on a local machine? maybe --no-ai along with something like a few thousand patch downloads?

what is the performance improvement/hit, if any? is there a noticeable win here?

OllieinCanada:
I attempted the requested real-machine workload with the exact PR head: a fresh database, current shallow Linux checkout, and `sashiko --download 2000 --no-ai --no-api`. The run was not valid performance evidence: while Sashiko was fetching maintainer remotes, the local WSL filesystem transitioned to read-only, the temporary SQLite database reported I/O errors and corruption, and I stopped the process. The partial 338.04 s result is therefore not comparable or publishable.

The concurrency regressions and CI are green, but I cannot substantiate a real end-to-end performance claim from this environment. I’m closing this voluntarily instead of leaving an unsupported optimization PR in the review queue. The atomic-outbox tests and implementation remain available here if the project later wants to revive the correctness change with maintainer-owned workload data.

PR #436, opened 2026-08-15
patchwork: support configurable user agent

### LINKED ISSUE #282: Check Patchwork API protection
Quoting @matttbe  from an off-list exchange wrt to Patchwork Integration

>  to avoid being blocked by AI protections, it is recommended to use a custom user agent
> (Don't use a non-valid email address, or one for a bot -- also, probably
> best to add a config for that, and not to commit the user agent in the
> code.)

This has yet to be proven/tested, but filing this here for tracking.

OllieinCanada:
Fixes #282

## Summary

Allow each Patchwork API policy to configure an identifiable HTTP
User-Agent.

## Why

Some Patchwork deployments use automated-client protections and may reject
requests that do not identify a real service contact. Sashiko previously
provided no way to set that header, so operators could not follow an
instance's client-identification policy without changing the code.

## Changes

- add optional patchwork.user_agent email-policy configuration
- resolve the setting at outbox delivery time alongside the API token
- send it on both the message-ID lookup and check creation requests
- document the setting and add it to the example policy

No default identity is hard-coded; each deployment controls the value it
publishes.

## Validation

- deterministic localhost test captures the real lookup and check requests
  and verifies their User-Agent and authorization headers
- cargo fmt --all -- --check passes
- git diff --check passes
- local Rust compilation is blocked before Sashiko builds because aws-lc-sys
  rejects the installed GCC 9.4 compiler; GitHub CI provides the authoritative
  clippy and unit-test results


derekbarbosa:
Hi Ollie, could you please fix those commits? IOW - have the SOBs and authorship match the account in which you are submitting this PR with? I'll review once that happens :)

OllieinCanada:
Confirmed the requested identity repair is present on the current head. Both commits are authored by Oliver Slapinski <olliefromcanada@gmail.com> and carry the matching Signed-off-by line; the DCO, lint, and unit-test checks are green. The branch is also cleanly mergeable. This is ready for your review when convenient.

PR #437, opened 2026-08-17
Add codebase scanning tool

FFizzZZ:
## Summary
- add TraeCLI as a review provider and expose structured review findings
- add a `tools/codebase-scan` package for whole-codebase scanning with inventory, grouping, synthetic patch generation, Sashiko orchestration, validation, and report rendering
- wire the tool into project docs, packaging, Makefile tests, and GitHub Actions

## Testing
- Not run in this PR creation step


rgushchin:
Thank you for the pull request! I really like the idea of adding a scanning tool, but I think it should be better integrated into the sashiko source code rather than a python wrapper on top of it. Wdyt about such approach?

FFizzZZ:
Thank you for the feedback! I'll rework the implementation to integrate the scanning tool directly into the Sashiko source code.

PR #419, opened 2026-08-11
Cherry-pick merge conflict review

helq:
On top of #418 and #413 

Add a dedicated review pipeline for cherry-pick merge conflicts, enabling Sashiko to review the correctness of conflict resolutions.

This cherry-pick review pipeline is separate of the review sashiko does and should never affect it. It has being built on top of #413, with a simple interface to define the graph that defines the review.

Commits from 3f8bfb6563f53705c41e4d5585215e8a3906df7e to b091bcad9bf11edd6912e8704b8cd1eae110330c

Some details on the implementation.

- Data model: add review_context column and ReviewKind enum to distinguish patch reviews from cherry-pick reviews.
- Review context types: define CherryPickReviewContext with finding filter for cherry-pick-specific analysis.
- Stage prompts: add 10 prompt templates for cherry-pick review stages (semantic intent, dropped changes, merge correctness, origin classification, etc.).
- CherryPickReviewPipeline: implement the pipeline as a principled Pipeline over the shared stage machinery, running cherry-pick-specific analysis stages alongside shared analysis, dedup, synthesis, and verification.
- Dispatch: route reviews to the cherry-pick pipeline when review_context indicates a cherry-pick.
- Daemon integration: read and forward review context from the daemon into the worker.
- API submit variant: add a cherry-pick submit endpoint that accepts supporting commits and enqueues via enqueue_fetch_with_support()

PR #450, opened 2026-08-21
Cherry-pick review testing

helq:
On top of #419 

The top two commits include a series of tests for the cherry-pick review feature. The main consideration is that we can run sashiko and gather a trace, and then we can use this trace to prevent some regressions while in development. This strategy helped me develop the whole pipeline infrastructure by first experimenting with sashiko and changing everything on it.


helq:
Moving to new v2 version soon

PR #452, opened 2026-08-22
db, api: Optimize database concurrency, query performance, and durability

helq:
On top of #450 

This change resolves query stalls and improves API responsiveness under heavy review workloads by optimizing SQLite concurrency, indexing, and connection management.

Previously, high worker write volume and unindexed aggregations caused read lock contention and slow patchset listing queries.

Changes overview:
  • Reader Pool: Added 4-connection reader pool to db.rs with lock-free round-robin borrowing for local SQLite databases.
  • PRAGMA Tuning: Configured journal_mode = WAL, synchronous = NORMAL, cache_size = -64MB, mmap_size = 256MB, and busy_timeout = 5000ms.
  • Periodic WAL Flusher: Background flusher task periodically triggers PRAGMA wal_checkpoint(PASSIVE) to control WAL file growth without blocking active readers.
  • CTE Query Refactoring: Scoped findings severity aggregation in db.rs to only the paginated patchset IDs.
  • Indexes & Migrations: Added idx_reviews_created_at and idx_findings_review_severity to schema.sql with dynamic idempotent startup migration.
  • Cache Bounding: Bounded api.rs to 500 entries with TTL/timestamp eviction and added ?nocache=true parameter support.


rgushchin:
Sorry, but I can't not see this as an example of how to not post pull requests. It's a huge PR combining a ton of different independent optimizations in arguably the most scary area without a clear statement of which problem it solves and any data to back up the changes. Maybe they all are perfect and strictly necessary but I don't know how should I evaluate it.
Please, focus on fixing a single problem at a time with a minimum changes required to solve this particular problem.

PR #457, opened 2026-08-25
workflow: restore three behaviours the stage session lost with call_tools

ftabba:
`StageSession` implements `call_tool` but not `call_tools`, so it inherits the `LlmSession` default. That default propagates a tool error with `?`, so any call the toolbox rejects ends the stage and with it the whole review. I hit it reviewing an arm64 patch: the model called `git_read_files` without `revision` and the run died with

    AI review for patch 1 failed with exception: Missing revision

after which `local_review` restarted the review from stage 0. It retries three times, and each attempt is a full multi-stage run over the whole patch and its gathered context, so a run can spend a great deal and produce nothing. It is intermittent, since it needs the model to emit a call the toolbox happens to reject.

Two more behaviours went the same way. The hand-rolled loop ran the batch through `join_all`; the default runs it in a sequential `for`. And it blocked a consecutive duplicate call with a synthetic error, which was added deliberately in "feat(worker): implement action history hashing loop prevention" to stop a model burning its turn budget.

The reason CI stayed green through all three: `ReviewStageSession` still has them, but it is reachable only through `src/pipelines`, which nothing in the tree calls, and the two unit tests covering duplicate blocking exercise that path rather than the engine.

One commit each, and every behavioural change carries a test I checked by breaking the fix under it. Removing the first commit's override makes its test fail with the same `Missing revision`. The third commit's tests fail if the guard is made per batch instead of per session, or if it compares only the tool name and not the arguments.

The second commit carries two tests, because one of them does not do the job on its own. It asserts the results stay in the order of the calls, which a sequential implementation also satisfies. That order is still worth asserting, because the Gemini path drops the call id on the wire and a tool result carries only the function name, so two calls to the same tool in one turn are told apart by position. Detecting sequential execution needs a tool that blocks until the rest of its batch arrives, so `ToolBox` grows a way to register one; it is behind `cfg(test)` and crate-visible, so the API is unchanged. Awaiting the batch one call at a time makes that test time out.

One thing I did not do. `StageSession::call_tools` and `ReviewStageSession::call_tools` are now near-identical, which the coding standards would rather they were not. The cleaner answer may be deleting `src/pipelines` and `ReviewStageSession` instead, since nothing calls them. Either is a separate change and your call, so I have left both alone.

### DIFF
--- a/src/ai/session.rs
+++ b/src/ai/session.rs
@@ -102,7 +102,10 @@ pub trait LlmSession: Send {
     }
 
     /// Executes multiple tool calls requested by the LLM.
-    /// Default implementation runs them sequentially.
+    /// Default implementation runs them sequentially, and propagates a tool
+    /// error rather than reporting it to the model, which ends the session.
+    /// A session that wants the model to see and correct its own bad calls
+    /// must override this.
     async fn call_tools(&mut self, calls: Vec<ToolCall>) -> Result<Vec<(String, Value)>> {
         let mut results = Vec::with_capacity(calls.len());
         for call in calls {
--- a/src/toolbox/mod.rs
+++ b/src/toolbox/mod.rs
@@ -103,6 +103,16 @@ impl ToolBox {
         }
     }
 
+    /// Registers an extra tool. Test-only: the review toolbox is fixed, and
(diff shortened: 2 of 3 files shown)


rgushchin:
Good catch, thanks!

PR #458, opened 2026-08-25
worker: restore four behaviours lost in the workflow migration

ftabba:
`designs/DESIGN_DECLARATIVE_AI_WORKFLOWS.md` sets the bar for the migration at "100% behavioral parity with existing review results". These are four places the engine differs from the hand-rolled loop it replaced. The first three change what the model is sent, the fourth what the progress bar is told. None changes the workflow graph.

**Stage 7 lost the commit message.** `use_log_in_context` on `ReviewStage` was true for stages 1, 2 and 7 and false for 3 to 6, as one expression. The workflow states it per stage as a boolean argument, and stage 7 is passed `false`, so it has been reviewing hardware and driver code from the diff hunks alone while its own instruction asks it to judge whether the patch is driver code at all. Rather than flip the one argument I derived the rule from the stage number and deleted the argument, so the seven call sites cannot disagree with each other again.

**Stages 1 to 7 lost five pieces of their instruction.** The old `format_guidance` told the model not to call `TodoWrite` and to carry a suspected bug from that checklist forward as a concern; said what to return when there is nothing to report; gave the meaning of each field; restricted `dismissed_concerns` to candidates investigated and disproved; and told it not to dismiss a concern by assuming the caller handles it. Eleven of the vendored prompts still mention `TodoWrite`, among them `callstack.md`, which stage 3 includes. Stage 10 kept an anti-charity directive of its own, which is why the loss from 1 to 7 reads as an oversight. This keeps the stricter key lists and the `line` and `code_snippet` fields the migration introduced.

**Included files render at the end of the prompt, not where the template puts them.** The design doc says a template's `@include("path")` directives "are expanded inline"; rendering appends instead. Two prompts already depend on the documented behaviour. The system prompt opens `<global_review_guidelines>`, says the documents that follow are the source of truth, and closes it again with nothing in between, while the guides arrive after the patch and the pre-fetched source. And the pre-screen prompt carries a literal `@include("subsystem/subsystem.md")` inside its `<subsystem_guide_index>` tags, which reaches the model as text while the index itself arrives after the patch. A file now renders at its directive when the template has one and is appended when it does not, so a template naming no place behaves as before. The pre-screen prompt needs no change; its directive starts working, and resolves to nothing when the index is absent rather than reaching the model as text. Directives are matched against the template and never against a variable's value, which matters because sashiko reviews arbitrary patches: a diff containing a directive as part of its own text would otherwise place a file into itself.

**The progress bar counts to eleven whatever the plan says.** `ReviewStarted` is emitted when the planner finishes, but the planner is skipped exactly when stages are chosen by hand, and a skipped stage returns before emitting anything, so the branch meant to report a hand-picked list is unreachable. When the planner does run its result is discarded for a hardcoded list of all eleven stages. The list is known where the fan-out resolves it, which happens either way, so it is emitted from there now.

Each commit stands alone and carries a test. A fifth fix belongs with these and is not here: the subsystem guides are dropped on a `--stages` run, so `--stages 1,2,5` sends the model none of the 67 guides under `third_party/prompts/kernel/subsystem/`. Skipping only the planner, as the old code did, breaks four existing tests, because Phase 0 also tolerated a missing `subsystem.md` and a failed response and the stage has neither tolerance, so that fix makes the pre-screen fatal where it used to degrade. How would you want a stage to fail when its failure should not end the review? Given that, this is a small fix on top and I am happy to write it.

### DIFF
--- a/src/worker/kernel_workflow.rs
+++ b/src/worker/kernel_workflow.rs
@@ -261,10 +261,23 @@ Follow the formatting rules strictly. Do not use markdown headers or ALL CAPS sh
 SPECIFICITY REQUIREMENT: Each inline comment MUST reference the exact function name, file, line number when known, and specific triggering condition. Prefer the finding's `locations` field when present. Do not produce vague summaries like 'potential issue in error handling'. State precisely what goes wrong, where, and under what circumstances. Do not invent line numbers; if the exact line is unavailable, anchor the comment to the nearest verified function or symbol and explain the triggering condition."#;
 
 const STAGE_JSON_SCHEMA_EXAMPLE: &str = r#"
-Return ONLY a JSON object with 'concerns' and 'dismissed_concerns' arrays.
+TodoWrite compatibility: vendored prompts may ask you to add tasks or suspected bugs to TodoWrite. Do not call or mention TodoWrite. Treat those instructions as an internal checklist only. If that checklist identifies a concrete suspected bug, carry it forward as a JSON concern with file, function_or_symbol, line when known, triggering condition, and evidence. Do not output generic checklist progress as a concern.
+
+Once you have gathered sufficient information, return ONLY a JSON object with 'concerns' and 'dismissed_concerns' arrays.
+If you find no concerns and no dismissed concerns, return {"concerns": [], "dismissed_concerns": []}.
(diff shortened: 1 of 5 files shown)


rgushchin:
All good fixes, thanks!

PR #472, opened 2026-08-31
nntp: Add TLS support for the news session

chucklever:
NntpClient speaks only cleartext. lore.kernel.org offers neither
STARTTLS nor a 563 listener, but an internal public-inbox mirror
can require TLS.

NNTPS is implicit TLS. The handshake runs before the greeting, so
every command method is untouched (patch 1). Trust anchors come
from the host certificate store, which retains the local CA as a
property of the deployment rather than of the config. The config
does name the rustls crypto provider: ring and aws-lc-rs both
reach this binary through reqwest and lettre, and neither installs
a process default.

The knob is nntp.tls, default off, so an existing Settings.toml
continues to connect in the clear (patch 2). The port is not
derived from it. NNTPS listens on 563 rather than 119, and
deriving one from the other would silently move a server an
operator had pinned.

Note: I thought our internal security policy might require the use
of NNTPS, so I implemented these patches. But it turns out only port
22 is open between our internal sashiko and public-inbox instances.
So the patches are untested. Feel free to drop them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,6 +118,45 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d902e3d592a523def97af8f317b08ce16b7ab854c1985a0c671e6f15cebc236"
 
 [[package]]
+name = "asn1-rs"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b7f43a50ac4fdca5df8e885c21b835997f0a1cdee65494a6847694a98652d9d8"
+dependencies = [
+ "asn1-rs-derive",
+ "asn1-rs-impl",
+ "displaydoc",
+ "nom 7.1.3",
+ "num-traits",
+ "rusticata-macros",
+ "thiserror 2.0.20",
+ "time",
+]
+
+[[package]]
+name = "asn1-rs-derive"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3109e49b1e4909e9db6515a30c633684d68cdeaa252f215214cb4fa1a5bfee2c"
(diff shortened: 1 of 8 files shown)


chucklever:
Rebased onto 54a74baa37f8 to clear the merge conflict. The pair
applies cleanly again.

Two notes on the rebase.

The conflict in src/nntp.rs was mechanical. Your
group_validates_matching_group_name and this series'
tls_article_round_trip both append to the end of the same test
module. Both are kept.

d798e62b38cb adds two NntpClient::connect() call sites, one in the
GROUP error path and one in the ARTICLE error path. This series
gives connect() a tls argument, so both sites needed updating.
They now pass settings.nntp.tls. Left passing false, a reconnect
after a failed GROUP or ARTICLE would drop to plaintext with
nntp.tls set, while the initial connection stayed on TLS.

The caveat in the description still holds. None of this has been
tested against a real NNTPS server.

chucklever:
Rebased onto 06ba23398718 to clear the merge conflict; the pair applies cleanly again. Three notes.

The Cargo.toml conflict was mechanical: upstream added rustix beside the two rustls crates and a dev-dependency block beside rcgen. Both sides are kept. rustls-native-certs was already in the lock through reqwest's platform verifier, so the only new build weight is rcgen and its dev-only parsers.

The nntp section gained a Default impl upstream when it became optional, so the tls field is now initialized there too.

Three small fixes to patch 1 from a review pass over the rebased code: the close_notify write in quit() now runs under the client timeout like every other I/O, since a peer that has stopped reading can stall it; the TCP connect takes a (host, port) tuple, so an IPv6 literal that ServerName accepts also addresses the socket; and a dead Arc import in the test module is gone.

The caveat in the description still holds. None of this has been tested against a real NNTPS server.

PR #475, opened 2026-09-02
ai: split OpenAI Responses and compatible providers

martinhoyer:
## Summary

Split the existing OpenAI integration into two distinct providers:

- openai uses OpenAI’s native /v1/responses endpoint.
- openai-compatible retains the Chat Completions implementation.

## Motivation

Using OpenAI API with gpt-5.6 models currently fails on:
```
{
  "error": {
    "message": "Unsupported value: 'temperature' does not support 0.0 with this model. Only the default (1) value is supported.",
    "type": "invalid_request_error",
    "param": "temperature",
    "code": "unsupported_value"
  }
}
```

Shoehorning special temperature handling into the existing compatible provider did not solve the underlying problem. Modern OpenAI reasoning and tool-calling workflows ultimately need the Responses API, while third-party OpenAI-compatible services generally continue to expose Chat Completions.
Splitting openai and openai-compatible therefore seemed like the most practical and maintainable approach.

## Changes

- Add a dedicated OpenAI Responses API client.
- Add [ai.openai] configuration, separate from [ai.openai_compat].
- Clearly reject legacy-only provider = "openai" configurations with migration guidance.
- Preserve and replay complete Responses output items, including encrypted reasoning state and original function-call identifiers.
- Classify retryable OpenAI errors correctly.
- Handle JSON mode requirements and cached-input token accounting.
- Update settings, examples, provider documentation, and design documentation.

Not sure if there should be separate design doc as well.


rgushchin:
> [!NOTE]
> *This review was generated with the assistance of an AI tool.*

Overall looks great and well-architected! A few quick observations and potential regression risks:

1. **Default `max_tokens` (4,096) may truncate reasoning**:
   In `src/ai/mod.rs`, omitting `[ai.openai].max_tokens` defaults to 4,096. On the Responses API, `max_output_tokens` bounds the *sum* of internal reasoning tokens and visible output. Reasoning models like `gpt-5.6` with `medium` effort can easily exhaust 4,096 tokens during the reasoning phase alone, causing premature `incomplete` (truncated) responses. Consider defaulting to `65536` for the `openai` provider.

2. **Handle `status: "failed"` in `translate_ai_response`**:
   `translate_ai_response` checks `resp.status == "incomplete"`, but if OpenAI returns a 200 OK with `status: "failed"` and an `error` object, it currently returns `Ok(AiResponse)` with empty content and tool calls. Propagating the error message directly would prevent confusing downstream `ValidationError`s.

3. **Global cache invalidation**:
   Bumping `CACHE_KEY_VERSION` in `src/ai/cache.rs` to `"sashiko-ai-cache-v2"` invalidates cached responses for *all* providers (Gemini, Claude, Bedrock, Ollama, etc.). Users running benchmarks or workloads with `response_cache = true` will experience a full cold-cache restart across all backends.

4. **`base_url` doc & normalization**:
   `docs/configuration.md` lists `[ai.openai].base_url` as "model-derived", but `default_base_url()` statically points to `https://api.openai.com/v1/responses`. Also, `OpenAiClient` does not normalize paths (e.g. appending `/responses`), so custom proxy URLs like `http://host/v1` without `/responses` will fail.

5. **`temperature` suppression**:
   `translate_ai_request` unconditionally passes `temperature: None`. This fixes `gpt-5.6` rejecting `0.0`, but prevents tuning temperature for non-reasoning models (like `gpt-4o`) or future models that support it on `/v1/responses`.

martinhoyer:
> Consider defaulting to `65536` for the `openai` provider.

Thanks, already noticed 4096 is too low. Trying 16k and it was quite enough for sol with xhigh reasoning for a patch review. Wouldn't 65k be an overkill?

martinhoyer:
@rgushchin adressing feedback in fixup commits, thanks.

- Raised the default output budget from 4K to 16K, validated with Sol/xhigh, and now log incomplete reasons and token usage.
- Surface failed Responses as provider errors instead of empty responses.
- Scoped cache invalidation to OpenAI Responses only.
- Corrected base_url documentation and added /responses normalization.
- Forward temperature for supported models while omitting it for known reasoning models.

rgushchin:
Can you please rebase it and also squash fixups into corresponding commits? Thanks

martinhoyer:
> Can you please rebase it and also squash fixups into corresponding commits? Thanks

@rgushchin Done :saluting_face:

martinhoyer:
@rgushchin Rebased once more. Please let me know if you want me to keep it in sync with main or I shouldn't bother.

rgushchin:
Sorry, was busy with merging some other stuff. Can you, please, check some findings here?
https://sashiko.sashiko.dev/#/patchset/sashiko-475

martinhoyer:
> Sorry, was busy with merging some other stuff. Can you, please, check some findings here? https://sashiko.sashiko.dev/#/patchset/sashiko-475

Thanks, addressed and rebased. I'm not sure about the "Missing validation in commit message" - did I miss some commit message requirements other than being signed?

martinhoyer:
@rgushchin I've addressed the v2 review and then started dogfooding `sashiko review` w/ openai provider. A lot of changes were needed, especially around preserving backwards compatibility.

martinhoyer:
> ### Sashiko review — v3

Not sure about this - I've let Claude review the feedback:

[HIGH] prune_payload window function / write lock contention — Not valid at this severity.

The claim says the window function "scans every row without a LIMIT" and will "severely degrade daemon throughput." But:

1. The cache has a hard cap of MAX_CACHE_PAYLOAD_BYTES = 256 MB (cache.rs:16). The row count is bounded by this cap divided by average entry size. A few hundred to low-thousands of rows is the realistic ceiling.
2. SQLite computes length(CAST(x AS BLOB)) in O(1) per row — it reads the byte length from the record header, not the blob content itself.
3. The window function (SUM ... OVER (ORDER BY ... ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)) is a single-pass running sum over the sorted set, not a nested loop. This is O(n) in the bounded row count.
4. There's already a write_lock: tokio::sync::Mutex (cache.rs:47) serializing writes, and WAL mode (cache.rs:57) prevents readers from blocking on writes. The only contention is write-vs-write, which the mutex already serializes.
5. busy_timeout = 5000 (cache.rs:62) covers the remaining edge case.

For a bounded table of this size, the entire prune runs in single-digit milliseconds. This is not a throughput bottleneck.

---

[MEDIUM] Sync JSON processing on Tokio worker threads — Technically correct, practically negligible.

The claim correctly identifies that decode_response, provider_metadata_size, and diagnostic_request_json do synchronous CPU work on the async runtime. However:

1. These all execute in the tail of an API call that takes seconds of network I/O. The CPU work (even at the 17 MB body limit) is tens of milliseconds.
2. Real-world payloads are typically a few KB to a few hundred KB — nowhere near the 17 MB hard limit, which exists for edge cases like large continuation metadata.
3. decode_response intentionally parses twice (budget count + deserialize) as defense against JSON complexity bombs (MAX_RESPONSE_JSON_VALUES = 128K at openai_responses.rs:50). Moving to spawn_blocking wouldn't remove the double parse.
4. The Tokio worker pool has multiple threads. One thread blocked for 20-50 ms during response handling (which just completed a multi-second API call anyway) doesn't meaningfully stall the cooperative scheduler in practice.

Adding spawn_blocking here would be correct for async hygiene in a generic library, but it wouldn't produce a measurable improvement in this daemon's throughput. The cure (extra thread hops, allocations for moving data to/from the blocking pool) may well cost more than the disease for typical payloads.

PR #478, opened 2026-09-03
e-mail: Send MTD reviews to all the recipients

miquelraynal:
Sending the reviews to the maintainers only applies more pressure on us, so let's just make these reviews fully public by hitting the "answer all" virtual button.

Link: https://lore.kernel.org/linux-mtd/87v78yimb1.fsf@bootlin.com/

### DIFF
--- a/sashiko.dev/email_policy.toml
+++ b/sashiko.dev/email_policy.toml
@@ -299,12 +299,8 @@ cc = ["media-ci@linuxtv.org"]
 
 [subsystems.mtd]
 lists = ["linux-mtd@lists.infradead.org"]
-cc = ["Miquel Raynal <miquel.raynal@bootlin.com>",
-      "Richard Weinberger <richard@nod.at>",
-      "Vignesh Raghavendra <vigneshr@ti.com>",
-      "Pratyush Yadav <pratyush@kernel.org>",
-      "Michael Walle <mwalle@kernel.org>",
-      "Takahiro Kuwano <takahiro.kuwano@infineon.com>"]
+reply_all = true
+reply_to_author = true
 
 [subsystems.audit]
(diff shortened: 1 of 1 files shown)


rgushchin:
Thanks!

On your comments in the thread:

> About Sashiko itself:
> + many interesting findings, worth enabling for sure
> + most of the findings are correct, but not always worth another version

Please, provide a feedback when the severity estimation is wrong and those are marked anything but low severity.

> - no possible feedback loop (eg. "this is an hallucination, stop
  bothering us with this")

I'm working on it, it's not trivial, but eventually we'll have something like this.

> - pre-existing issues are flagged at the same "level" as new issues

This will be changed soon.

miquelraynal:
Hi @rgushchin, thanks a lot for the feedback, very appreciated. I am super excited to know that you work on getting Sashiko capable of feedback loops, this will be very interesting.

> Please, provide a feedback when the severity estimation is wrong and those are marked anything but low severity.

First case I met (on clocks) after your comment is [this one](https://lore.kernel.org/linux-clk/20260813-perso-ma35d1-upstream-clk-v1-1-e78e5e6172ea@bootlin.com/T/#mb4f8c5cf64e89c5b27fac5b07a6770f088aadc34).

Severity is high. Sashiko sees a possible error pointer dereference, that would be a crash, yes indeed. But how likely is this? The only possibility is a clock registration returning -ENOMEM very early at boot time. This is purely theoretical and cannot happen in practice (or one would arguably have bigger problems). This is pure theoretical hardening, nothing big, hence it should -IMHO- get a low severity.

Anyway, thanks a lot for working on this tool, it is definitely useful, looking forward to see it evolve!

Cheers,
Miquèl

PR #484, opened 2026-09-05
review: point read_prompt at the prompt directory, not a file

kees:
The read_prompt tool resolves the name it is given against a base path, so that base has to be the prompt directory: ToolBox is constructed with one everywhere else, and its test reads "technical-patterns.md" relative to it. Worker-run reviews instead passed `<prompts>/tool.md`, a file, and one that exists in no prompt bundle. Every call therefore failed while canonicalizing the base, with "Failed to canonicalize base path: No such file or directory".

The tool is how a stage reaches a guide the pre-screen did not select: any of the subsystem guides, severity.md, false-positive-guide.md. Losing it costs a review the guidance it went looking for, and it fails quietly, because a tool erroring is indistinguishable from a model deciding not to call it.

Pass the directory. The test pins both directions, since the failure only shows up as an absence.

### DIFF
--- a/src/local_review.rs
+++ b/src/local_review.rs
@@ -509,7 +509,8 @@ async fn review_single_patch(
         let provider =
             crate::ai::create_provider_from_ai(ai).context("Failed to create AI provider")?;
         let provider = decorate_provider(provider, ai, llm_semaphore, quota, &retry_budget);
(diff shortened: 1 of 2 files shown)


rgushchin:
Idk if we really need the read_prompt tool anymore, maybe it's better to drop it. But good catch anyways, thanks!

PR #485, opened 2026-09-05
Add "testing" review stage

kees:
This adds a missing review stage persona: "testing". There are a few interesting pieces here, and I'm not sure the right way to approach them:

1) The style guides need to be included in the prompt verbatim, and asking the LLMs nicely to read the files didn't seem to work, so I explicitly inject their contents. This creates a dependency on the harness for the prompts if they get moved to the upstream prompts collection.

2) I had to renumber everything. I *think* this is okay, but wow was it a lot of churn. Maybe the stages need to be refactored in some way to avoid the explicit numbering?

What do folks think about this approach? It worked well for a few series I tested it on.


kees:
I've got a much better series coming. Closing this in favor of that one.

PR #487, opened 2026-09-05
prompts: clarify report's commit header format

kees:
While watching stage turns, we noticed sometimes the report generation stage would make multiple needless round-trips with the sanity checker because it kept producing a header starting with "commit:" instead of the checked "commit ". This is a weakness of the report prompt.

The inline report must open with "commit <hash>", and the validator checks for exactly that: the literal prefix "commit " at the start of a line. The template's Structure list, which is where the report's contents are specified, asked only for the "git sha of the commit" and immediately below it for the "Author: line". Content but not form, with a colon shown in the neighbouring item and more colons in the Link: and Date: lines nearby. So "commit:" is a reasonable guess, and a report that guesses it is rejected and regenerated in full.

Specify the header where the list already describes it: the word commit, one space, the sha, no colon, naming the wrong forms. Say why it differs from the lines under it, since that should help specify it well.


kees:
This change is stand-alone, so I think it's actually better if I send this upstream to `review-prompts`. I've done it now: https://github.com/masoncl/review-prompts/pull/99

rgushchin:
Ok, I'll pull it from Chris's repo once merged there. Closing here. Please feel free to re-open if you want to merge it directly. Also, please don't bump the revision, it's not a 1:1 match anyway.

PR #488, opened 2026-09-07
Add "testability" stage

kees:
This adds a missing review stage persona: "testability". It contains a bunch of refactoring work before the new stage is added:

- move from numbered stages to named stages
- drive more of the stage logic from a common table
- deal with impact on --stages

Note that this also depends on getting the new testing guides from review-prompts PR: https://github.com/masoncl/review-prompts/pull/100


kees:
Cc @justinstitt @bwendling

JustinStitt (approved):
hi kees

this lgtm and [`b614236`](https://github.com/sashiko-dev/sashiko/pull/488/commits/b614236a7629089fd970a4c4d4ad64488c789502) is a nice touch as well.

bwendling (commented):
My Rust-fu isn't great, so I assume that the changes work. :-)

My one comment is whether it would be better to split out the NFC refactoring changes into a separate PR and then add the "testability" stage?

kees:
> My one comment is whether it would be better to split out the NFC refactoring changes into a separate PR and then add the "testability" stage?

I split the changes up in logic steps, 1 per commit, so I think this is already cleanly addressed. I think if the PR isn't squash-applied, it should show up cleanly in the commit history.

bwendling:
> > My one comment is whether it would be better to split out the NFC refactoring changes into a separate PR and then add the "testability" stage?
> 
> I split the changes up in logic steps, 1 per commit, so I think this is already cleanly addressed. I think if the PR isn't squash-applied, it should show up cleanly in the commit history.

Ah! Gotcha...I resind my remark. :-)

rgushchin:
Directionally I like it but there are drastically different views on testing in various subsystems (which is a separate problem), so I wonder how much noise and hostility it might create. I wonder if we can start with one or two subsystems which will opt-in?

kees:
The prompts as written should be very very conservative. The primary thing it wires up is style checking for kunit and kselftest. What do you think of the general number->name refactor? Should I split that into a separate PR?

rgushchin:
I like number->name refactor. It makes total sense, especially as we get more stages which can be run in parallel and using various models and also as we get more workflows. Yes, if possible, please, separate it.
On testing I'd need a bit more time to run it myself on various patchsets - I'm still a bit worried about how it might work. Or at least it should be released separately so we can easily roll it back. Sorry for being too conservative here - but it's so easy to polarize the community with anything ai* :)

PR #492, opened 2026-09-07
ai: optionally refuse a tool call already made in the same stage

shardulsdk-mpiric:
`StageSession` blocks a tool call that repeats the one immediately before
it, so a model that cycles between several calls is never caught. Counting
identical name and argument pairs across the recorded turns of 28 local
review runs here, the share of calls repeating an earlier one reached 48
percent, and one run issued the same `git_read_files` call 69 times. Every
repeat costs a turn, and every turn resends the whole conversation, so the
waste compounds rather than staying flat.

The second commit adds `ai.dedup_tool_calls`. When set, a stage remembers
every call it has run and refuses an exact repeat, naming the earlier call
that already produced the answer and pointing at its result above in the
conversation, so the model has somewhere to go rather than only being told
no. It defaults to false, so nothing changes unless it is switched on, and
`test_non_consecutive_duplicate_tool_call_runs`, which asserts the current
behaviour, is untouched and still passes.

The first commit is a prerequisite rather than a cleanup. The guard existed
twice, in `call_tool` and again in `call_tools`, with the comparison, the
log line and the refusal text written out separately in each. `call_tools`
is the path a provider that batches its calls actually takes. I found this
the hard way: I first added the new check to `call_tool` alone, the whole
suite passed, and the setting did nothing at all, because the batched path
has its own copy. Sharing the two through `duplicate_refusal` and
`record_tool_call` means a change to the rule cannot reach one path and
miss the other.

`make check-all` passes. Both commits are covered by tests, including a
mirror of the existing non-consecutive test that asserts the repeat is
refused when the setting is on, and that the refusal names the earlier
call.

What I have not done: measured the effect on review quality or cost with
the setting enabled. The rates above count repeats in runs that were
recorded for other reasons, not a controlled comparison of reviews with
and without it, so I cannot say what it saves in practice beyond the turns
it removes. They are also whole-run figures rather than per stage, since
the turn records carry no stage boundary.

### DIFF
--- a/src/local_review.rs
+++ b/src/local_review.rs
@@ -559,6 +559,7 @@ async fn review_single_patch(
                 max_input_tokens: ai.max_input_tokens,
                 max_interactions: ai.max_interactions,
                 temperature: ai.temperature,
+                dedup_tool_calls: ai.dedup_tool_calls,
                 custom_prompt: options.custom_prompt.clone(),
                 series_range,
                 baseline_sha: Some(baseline_sha.to_string()),
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -365,6 +365,18 @@ pub struct AiSettings {
     /// Useful for debugging but verbose; disabled by default.
     #[serde(default)]
     pub log_turns: bool,
+    /// Refuse a tool call already made anywhere in the same stage, rather than
+    /// only the one immediately before it.
+    ///
+    /// The existing guard compares against the previous call alone, so a model
+    /// cycling between several calls is never caught. Measured repeat rates
+    /// within a single stage session were 53, 59 and 74 percent, with one
+    /// identical git_grep issued nine times. Every repeat costs a turn, and
+    /// every turn resends the conversation.
+    ///
+    /// Defaults to false, which preserves the existing behaviour.
+    #[serde(default)]
+    pub dedup_tool_calls: bool,
     #[serde(default)]
     pub response_cache: bool,
     #[serde(default = "default_response_cache_ttl_days")]
--- a/src/worker/prompts.rs
+++ b/src/worker/prompts.rs
@@ -85,6 +85,8 @@ pub struct WorkerConfig {
     pub max_input_tokens: usize,
     pub max_interactions: usize,
     pub temperature: f32,
+    /// See AiSettings::dedup_tool_calls.
+    pub dedup_tool_calls: bool,
     pub custom_prompt: Option<String>,
     pub series_range: Option<String>,
     pub baseline_sha: Option<String>,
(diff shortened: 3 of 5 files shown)


rgushchin:
Thanks!
2c from me:
1) I don't like adding another config option for this, unless there is a really strong reason for it. Let's figure out what is causing LLMs to repeat calls and maybe fix that instead. Can you share specific reviews with many repeated calls?
2) As you said, there is no data regarding the review quality, which is concerning.

rgushchin:
Sorry, merged by a mistake and had to revert it. Please, look into comments above.

shardulsdk-mpiric:
Sorry for the slow reply. Both of your points were fair, and when I went and measured what you asked about, the data did not support this PR.

The numbers behind it came from local runs with open-weight models through an OpenAI-compatible endpoint (qwen3-coder, glm-4.5-air and similar). Those models appeared to repeat calls often, but they also had other problems that make them poor Sashiko backends in practice, so they are not a good reason to change behavior for everyone. Those figures were also counted over whole runs, not per stage, so I would not rely on them.

To get data you can check, I added per-stage tool call tracing on a branch [1] and reviewed one kernel patch ("mptcp: do not drop partial packets") with gemini-3.1-pro-preview and the default [ai] settings. Across six analysis stages the model made 92 tool calls. Four repeated an earlier call in the same
stage:

- security asked for the same git_grep on two consecutive turns, twice. The existing guard refused the repeat each time, and the model moved on to other calls.
- resources re-read two file ranges it had viewed 4 to 5 calls earlier. The guard does not catch those, and the ToolBox cache answered them.

That is one patch, so it is a single data point, not a rate. But it is the opposite of the picture that motivated this PR: on the default model, repeats were rare and the model did not loop. I do not think the change is worth making, and I am not pursuing it.

The testing branch only adds recording: the stage name in the log tag, and one JSON line per tool call with its stage, turn, arguments, and whether the guard refused it or the cache answered it. The review itself, guard included, is upstream's. The commit, settings, full log and trace, with a README on how it was run, are in [1]. I am happy to run it over more patches if a real rate would be useful.

Two things turned up along the way that are probably worth more of your time than this PR was. I will send them separately:

- Token accounting leaves out Gemini's thinking tokens. The provider takes completion_tokens from candidatesTokenCount only, so on this review Sashiko reported 13.8k output tokens where the per-turn totals add up to about 152k. I have a small fix with a test.
- With response_cache on, a retried patch review replays the cached responses of the attempt that failed, and fails the same way. I will open an issue with a reproduction.

Thanks for the review.

[1] https://github.com/shardulsdk-mpiric/sashiko/tree/tooltrace-2026-09-24

PR #493, opened 2026-09-07
ai: honour response_cache for local reviews

shardulsdk-mpiric:
`AiSettings` carries `response_cache`, but only the daemon ever consulted
it. `sashiko review` builds its provider with `create_provider_from_ai`,
which is uncached, so turning the setting on did nothing on that path and
there was nothing in the output to say so.

`create_provider_cached` cannot be reused there, since it derives the cache
location from `settings.database.url` and a local review holds `AiSettings`
alone with no database beside it. This adds `create_provider_cached_from_ai`,
which places the cache under the XDG data directory already used for the
prompt bundle and lets `SASHIKO_RESPONSE_CACHE` override it. The helper that
resolves that directory is now shared rather than copied. `response_cache`
still defaults to false, so nothing changes unless it is switched on.

It matters most on a retry. A run that fails part way re-issues every
earlier call, and with no cache each of those is paid for again. One such
retry after a late-stage failure cost me $3.33 against a hosted endpoint,
which is what sent me looking at this.

`make check-all` passes, and there is a test for the path resolution
covering both the environment override and the XDG default. I am relying on
#480 having made the cache key safe across provider configurations rather
than having exercised that myself.

### DIFF
--- a/src/ai/mod.rs
+++ b/src/ai/mod.rs
@@ -421,6 +421,44 @@ pub async fn create_provider_cached(
     }
 }
 
+/// The response cache used by callers that have no database to sit beside.
+///
+/// `create_provider_cached` derives its location from `settings.database.url`,
+/// which only the daemon has. Local reviews get an XDG path instead,
+/// overridable with SASHIKO_RESPONSE_CACHE.
+pub fn local_response_cache_path() -> Result<std::path::PathBuf> {
+    if let Some(path) = std::env::var_os("SASHIKO_RESPONSE_CACHE") {
+        return Ok(std::path::PathBuf::from(path));
+    }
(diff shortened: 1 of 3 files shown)


kees:
I could use this! :)

I think it should use a common method, though. This is kind of duplicating code from the daemon. What about consolidating it like attached?

Also, I think the original PR suffered from creating two caches since the stdio workers under the daemon would end up creating the non-daemon cache location.
[0001-ai-honour-response_cache-for-local-reviews.patch](https://github.com/user-attachments/files/31978433/0001-ai-honour-response_cache-for-local-reviews.patch)

kees:
Updated version now opened as PR #520

PR #496, opened 2026-09-08
sync: Summarize what a sync cycle reached

chucklever:
A sync cycle that reaches no remote at all closes with "Sync cycle
complete", the same line a healthy cycle writes. The per-remote
failures are in the journal, three lines each, but the state they add
up to is recorded nowhere. One minute of a network outage put 116
lines in the journal and named no condition.

Counting failed returns would not name it either. A fetch that fails
against a remote whose local ref is recent enough falls back to that
ref and returns success, which is the path most remotes take when the
network is down. So ensure_remote now reports what happened rather
than whether it succeeded, and the cycle closes on a count of those
outcomes.

Two choices in the diff are worth a look. A remote nothing went out
to, because its fetch interval had not elapsed or a recent failure is
backing off, is counted apart from the ones that answered. The backoff
used to return an error, which would have made the cycle after an
outage report every remote as failed on a network that had recovered.
And the fallback path carries git's message out with it, because a
cycle where every remote fell back has no failed return to quote.

A cycle that tries more than one remote and fetches none logs at error
with the first failure. One remote tried and lost is that remote's
condition, and its own line already names it.

Verified standalone on top of 06ba23398718: sign-off valid, clippy
clean, 793 unit and 43 integration tests passing.

### DIFF
--- a/src/git_ops.rs
+++ b/src/git_ops.rs
@@ -849,12 +849,51 @@ async fn fetch_remote(
     }
 }
 
+/// What a remote's refs rest on once ensure_remote() returns.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum FetchOutcome {
+    /// The remote answered.
+    Fetched,
+    /// The fetch interval had not elapsed, so nothing was asked of
+    /// the remote and the refs are the ones already on disk.
+    Skipped,
+    /// A fetch failed within the last hour, so nothing was asked of
+    /// the remote and the refs are the ones the previous fetch
+    /// wrote.  A remote whose refs are missing or older than the
+    /// fallback window returns a BackoffError instead.
+    BackedOff,
+    /// The fetch failed and the refs left behind are the ones the
+    /// previous fetch wrote.  Carries what git said about the
+    /// failure.
+    StaleLocalRef(String),
+}
+
+/// A fetch failed within the last hour and the local refs cannot
+/// stand in for it, so backing off leaves the caller nothing to
+/// use.  Nothing was asked of the remote.
+#[derive(Debug)]
+pub struct BackoffError {
+    name: String,
+}
+
+impl std::fmt::Display for BackoffError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Remote {} failed recently and has no usable local ref, backing off",
+            self.name
+        )
+    }
+}
+
+impl std::error::Error for BackoffError {}
+
 pub async fn ensure_remote(
     repo_path: &Path,
     name: &str,
     url: &str,
     force_fetch: bool,
-) -> Result<()> {
+) -> Result<FetchOutcome> {
     // 1. Validate repo_path to prevent git from traversing up to parent repos
     if !repo_path.join(".git").exists() && !repo_path.join("HEAD").exists() {
         return Err(anyhow::anyhow!(
@@ -942,15 +981,6 @@ pub async fn ensure_remote(
         None => false,
     };
 
-    if failed_recently && !force_fetch {
-        let reason = "failed recently, backing off";
-        info!("Skipping fetch for {} ({})", name, reason);
-        return Err(anyhow::anyhow!(
-            "Remote {} failed recently, backing off",
-            name
-        ));
-    }
-
     // Check if HEAD exists
     let head_ref = format!("refs/remotes/{}/HEAD", name);
     let head_exists = Command::new("git")
@@ -960,6 +990,27 @@ pub async fn ensure_remote(
         .await
         .map(|s| s.success())
         .unwrap_or(false);
+
+    // The local ref stands in for a fetch that fails, or is backed
+    // off, only while it is recent enough.
+    let max_stale_age = std::time::Duration::from_secs(86400); // 24 hours
+    let can_fallback = head_exists
+        && match age {
+            Some(a) => a < max_stale_age,
+            None => false,
+        };
+
+    if failed_recently && !force_fetch {
+        if !can_fallback {
+            return Err(BackoffError {
+                name: name.to_string(),
+            }
+            .into());
+        }
+        let reason = "failed recently, backing off";
+        info!("Skipping fetch for {} ({})", name, reason);
+        return Ok(FetchOutcome::BackedOff);
+    }
 
     let should_fetch = if just_added || !head_exists || force_fetch {
         true
@@ -982,7 +1033,7 @@ pub async fn ensure_remote(
             "fresh"
         };
         info!("Skipping fetch for {} ({})", name, reason);
-        return Ok(());
+        return Ok(FetchOutcome::Skipped);
     }
 
     let mut fetch_ok = false;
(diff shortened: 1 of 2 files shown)


rgushchin:
> [!NOTE]
> *This review was prepared with the assistance of an AI code review pair.*

Hi @chucklever, thanks for this PR! The motivation makes total sense—having the sync worker report a clean completion when network outages cause remotes to silently fall back to stale local refs can be very misleading in production.

During a detailed review of the diff, a few issues came up (including one security consideration regarding credential redaction) that would be great to address:

---

### 1. Security: Credential Sanitization in `first_failure` (`src/worker/sync.rs`)

In `GitSyncWorker::run`, `first_failure` captures raw git stderr from `url_output.stderr`, `FetchOutcome::StaleLocalRef(message)`, and `Err(e)`:
```rust
if first_failure.is_empty() {
    first_failure = message;
}
```
When git fetches fail against authenticated HTTPS remotes (e.g. URLs with embedded personal access tokens or passwords), git's stderr frequently echoes the remote URL with the embedded token in the error string.

In `src/git_ops.rs` and `src/reviewer.rs`, git error strings are passed through `crate::utils::redact_secret()`. We should do the same here before formatting into the summary log:
```rust
let sanitized_failure = crate::utils::redact_secret(&first_failure);
```
Additionally, git stderr often contains embedded newlines (`fatal: ...\nerror: ...`). Replacing newlines with spaces or semicolons when printing the summary line will prevent log-splitting / multi-line pollution in log aggregators that expect single-line events.

---

### 2. Single-Remote Setups Never Trigger Error Summary (`tried > 1`)

The error condition for the cycle summary is:
```rust
let tried = fetched + stale + failed;
if fetched == 0 && tried > 1 {
    error!(...);
} else {
    info!(...);
}
```
In default / single-remote installations (e.g. standard local setups tracking only `origin` in `third_party/linux`), `tried` can never exceed 1. If that single remote fails or falls back to stale refs, `tried > 1` is never satisfied, and the worker will always log an `info!` level clean completion:
```
info!("GitSyncWorker: Sync cycle complete: 0 fetched, 0 skipped, 1 stale, 0 failed.")
```
Consider adjusting the condition to accommodate single-remote setups when failures occur, for example:
```rust
if fetched == 0 && tried > 0 && (tried > 1 || remotes.len() == 1) {
```
(Or logging at `warn!` if `failed > 0` regardless of `tried`).

---

### 3. Cycle Health Oscillation: `BackedOff` Grouped into `Skipped`

In `sync.rs`:
```rust
Ok(FetchOutcome::Skipped) | Ok(FetchOutcome::BackedOff) => skipped += 1,
```
- `Skipped` indicates a healthy remote whose sync interval has not yet elapsed.
- `BackedOff` indicates the remote recently failed and is in a backoff cooldown.

Because `BackedOff` is grouped under `skipped`, `tried` (defined as `fetched + stale + failed`) does not count it. During a persistent outage:
1. **Cycle 1:** Fetches fail/stale &rarr; logs `error!` with `N stale / failed`.
2. **Cycle 2 (1h later):** All remotes are backed off &rarr; counted as `skipped` &rarr; `tried = 0, fetched = 0` &rarr; logs `info!("... 0 fetched, N skipped, 0 stale, 0 failed.")`.

The worker oscillates between error alerts and completely clean green summaries every other cycle. Tracking `backed_off` as its own counter and including it in cycle health evaluation would prevent this flapping.

---

### 4. `ensure_remote` Returning `Ok(BackedOff)` When Local Ref Does Not Exist (`src/git_ops.rs`)

In `src/git_ops.rs`:
```rust
if !force_fetch && failed_recently(&repo_path, name).await? {
    info!("Skipping fetch for {} (backed off after recent failure)", name);
    return Ok(FetchOutcome::BackedOff);
}

// Check if HEAD exists
let head_ref = format!("refs/remotes/{}/HEAD", name);
let head_exists = ...;
```
`failed_recently` is checked before verifying `head_exists`. If a newly configured remote fails its very first fetch, it has no local ref. On subsequent non-forced calls, `ensure_remote()` returns `Ok(FetchOutcome::BackedOff)`.

Other callers of `ensure_remote` (such as `reviewer.rs:822`) check `if let Err(e) = ensure_remote(...)`. Because it now returns `Ok`, callers assume the remote is ready for use and proceed to query `refs/remotes/{name}/HEAD`, causing downstream failures with ambiguous argument errors.

`ensure_remote` should only return `Ok(FetchOutcome::BackedOff)` if `head_exists` is true; otherwise, if the local ref does not exist, it should return an `Err`.

chucklever:
Fixed all four in the updated branch.

1. The first failure now goes through `redact_secret()` with its lines joined into one, at the point it is captured. The per-remote lines in `git_ops.rs` log the same text unredacted and predate this PR; that is a separate change.
2. The guard is now `fetched == 0 && tried > 0`. The `remotes.len() == 1` form makes the level depend on the config rather than on what happened, and the plain form covers the single-remote case.
3. `BackedOff` has its own counter, printed on every summary. A cycle where every remote is backing off has no failure to quote and the network may have recovered since, so it logs at warn as "fetched nothing" rather than at error or as a clean cycle.
4. `ensure_remote()` checks for a usable local ref before honoring the backoff, and returns a typed `BackoffError` when there is none, so callers that only test `is_err()` bail as they did before this PR. The sync worker recognizes that error and counts the remote as backing off rather than failed. The backoff test now expects the error.

chucklever:
Rebased onto 06ba23398718 to clear the merge conflict. Two notes.

The conflict was in the git_ops.rs test module, where the local-branch-sync test and this patch's ensure_remote test were added at the same spot. Both are kept. The test now spawns git through git_cmd::in_dir_async, since 13af119be75d forbids a bare Command::new("git") outside git_cmd.rs.

f0411100619f redacts fetch_remote's stderr upstream, which covers the git_ops.rs half of item 1 from your review. The redaction at the point of capture in sync.rs stays, because the remote URL update path reaches the summary line through the same variable and is not covered by it.

Sign-off valid, clippy clean, 793 unit and 43 integration tests passing. The rework from 2026-09-09 that addressed your four items is otherwise unchanged; is there anything else you would like to see before this goes in?

chucklever:
Force-pushed one more revision, this one to the design.

The outcome-reporting version of ensure_remote() changed what the reviewer and the baseline lookup got back: a remote in backoff with a local ref under a day old became an Ok return, so a review would have proceeded against refs it did not ask to be refreshed, with nothing to say so. My 2026-09-09 note that those callers bail as before was true only when no usable ref existed.

The reporting now lives in a new refresh_remote(), which the sync worker calls. ensure_remote() wraps it and keeps its contract: every backoff is an error, whether or not a local ref could stand in, and a fetch that fails against a recent enough local ref succeeds as it always has. Commit message updated to say so.

Sign-off valid, clippy clean, 793 unit and 43 integration tests passing.

chucklever:
Force-pushed one more revision for the bot's v3 finding.

The get-url failure path logged git's stderr raw and redacted only the copy kept for the cycle summary. That stderr does not carry the URL in practice, since get-url fails only when the remote is gone or the config will not parse, but the code contradicted its own comment two lines up. The stderr is now redacted once, at capture, so both the warning and the summary see the same text.

Sign-off valid, clippy clean, 793 unit and 43 integration tests passing.

PR #497, opened 2026-09-09
Refactor numbered stages into named stages

kees:
Here is the numbered stages to named stages refactor commits, lifted from PR #488


rgushchin:
Merged manually. Thanks

PR #499, opened 2026-09-10
api: resolve patchset summary by slug as fallback

zraza-xlnx:
The GET /api/patchset summary handler only looked up patchsets by numeric id or message-id. Forge ingested patchsets are addressed by slug (for example reponame-79), so the web UI's summary requests for those patchsets fell through to the message-id lookup, matched nothing, and returned 404.

UI links reviews by slug, but the API never resolved slugs. This PR adds that fallback so slug links stop 404 error. 
https://github.com/sashiko-dev/sashiko/blob/39f6ce95c797bb40023247916a8b16d0f4aaf0da/static/index.html#L1967

Chain the non numeric lookups by result: try message-id first, then fall back to slug when the message-id lookup finds nothing. This mirrors the branching already used by the patchset details handler.

### Regression test results

| Test area | Example | Parent | PR |
| --- | --- | --- | --- |
| Repository-name and legacy slug lookup | `id=linux-next-725` (reponame-prnum) | FAIL: HTTP 404 | PASS: HTTP 200 |
| Numeric/message-ID lookup | `id=42` selects numeric patchset 42 | PASS | PASS |

This change is backward compatible.


zraza-xlnx:
Reviewed by @pieterj-xilinx and @alexaust-xlnx internally.

rgushchin:
Any reasons why it's closed? Thanks

zraza-xlnx:
Hi, this has been fixed in one of the other commits. https://github.com/sashiko-dev/sashiko/blob/68085717b3136625989d55d73c5318eb5bc5c1f7/src/api.rs#L1047
Thanks!

PR #500, opened 2026-09-10
ai/gemini: use 'user' role for function responses

Flowdalic:
When returning function/tool execution outputs to Google's Gemini API (generateContent):, setting the top-level Content role to 'function' causes the endpoint to reject the request with a 400 Bad Request error:

    Role 'function' is not supported. Please use a valid role: SYSTEM,
    USER, ASSISTANT, DEVELOPER, CONTEXT, USER_CONTEXT, MODEL...

In Gemini's API schema, function/tool responses containing a functionResponse part must be submitted under the 'user' role instead of OpenAI's convention of 'function' or 'tool'.

Map function response turns to the 'user' role when constructing the Gemini payload to satisfy API schema validation.

### DIFF
--- a/src/ai/gemini.rs
+++ b/src/ai/gemini.rs
@@ -570,9 +570,8 @@ fn translate_ai_request(request: AiRequest) -> Result<GenerateContentRequest> {
                 });
             }
             AiRole::Tool => {
-                // Gemini expects a 'function' role for tool responses
                 contents.push(Content {
-                    role: "function".to_string(),
+                    role: "user".to_string(),
                     parts: vec![Part::FunctionResponse {
                         function_response: FunctionResponse {
                             name: msg


src/ai/gemini.rs
zraza-xlnx, line 1056: Thanks! The role change seems correct.
One blocker: two existing tests still assert the old value and aren't updated, so the suite fails.

**Reproduce the error**:
cargo test --lib ai::gemini::tests::test_translate_ai_request

**Results**:
test result: FAILED. 3 passed; 2 failed
  at src/ai/gemini.rs:1056
  at src/ai/gemini.rs:1155

 **Fix : flip both assertions to "user"**
[sashiko/src/ai/gemini.rs at 1a07cfd951f18191d7776ff123a3abc908101207 · sashiko-dev/sashiko](https://github.com/sashiko-dev/sashiko/blob/1a07cfd951f18191d7776ff123a3abc908101207/src/ai/gemini.rs#L1056)
assert_eq!(gemini_req.contents[0].role, "user");

[sashiko/src/ai/gemini.rs at 1a07cfd951f18191d7776ff123a3abc908101207 · sashiko-dev/sashiko](https://github.com/sashiko-dev/sashiko/blob/1a07cfd951f18191d7776ff123a3abc908101207/src/ai/gemini.rs#L1155)
assert_eq!(gemini_req.contents[2].role, "user");

PR #503, opened 2026-09-12
email: Add a local sendmail transport

chucklever:
The email worker reaches a submission service through lettre's
relay(), which opens the session with implicit TLS. A submission
service that offers cleartext port 25 and nothing else is
unreachable, so a host whose only route out is a local relay cannot
send review mail at all.

The alternative a reviewer might propose is a knob that permits
plaintext SMTP. That knob cannot tell a loopback relay from a remote
one, so a later change of server would ship credentials and review
text in the clear across a network. The design here adds a transport
selector instead. Set smtp.transport to "sendmail" and the message is
piped to the local MTA, which leaves transport security and queueing
to the host mail configuration. The envelope goes on the sendmail
command line rather than through -t, so the recipients are the ones
sashiko addressed rather than whatever the MTA parses back out of the
headers.

Two consequences are worth checking. The key defaults to "smtp", so
an existing Settings.toml keeps parsing, but server and port mean
nothing to sendmail and become optional. The check that the smtp
transport has them moves to startup, and credentials supplied
alongside the sendmail transport are rejected there too. And a
message the MTA accepts is queued, not delivered. sendmail exits zero
at that point, so the outbox records the row as sent and a rejection
further along arrives as a bounce the outbox never sees.

Message construction is split out of send_email() first (patch 1) so
the second delivery path has somewhere to go other than the bottom of
a function already past the project's length guidance.

Cargo.lock grows by the async-std packages. Enabling lettre's
sendmail-transport feature records them through a weak feature edge
that nothing here turns on, so none of them are built.

Both patches were verified standalone on top of d0e8bfce601a:
sign-offs valid, clippy clean, 808 unit and 43 integration tests
passing.


chucklever:
Rebased onto 87980a1c0ea6 and force-pushed. Changes since the previous revision:

- The From header now goes through parse_lenient() like the other addresses, so a display-name sender address parses the same way in both transports.
- The dry-run path picks up the sign-in link withholding that landed upstream in the meantime.
- Cargo.lock reflects the current upstream versions of the async-std packages the sendmail-transport feature records.

Commit messages are unchanged. Sign-offs valid, clippy clean, 698 unit and 39 integration tests passing.

chucklever:
Rebased onto 06ba23398718 and force-pushed. The code is unchanged apart from context: the MailTransport block in settings.rs now sits after the NntpSettings Default impl that landed upstream. Cargo.lock moves the async-std packages the sendmail-transport feature records onto rustix 1.1.5, the version upstream now pins, so the lock resolves without a second rustix.

Sign-offs valid, clippy clean, 802 unit and 43 integration tests passing.

chucklever:
CI caught a race in test_sendmail_reports_a_nonzero_exit: the fake sendmail exited before lettre had written the message to its stdin, so the test saw EPIPE instead of the exit status. The script now drains stdin before rejecting, as an MTA does. Test-only change; force-pushed.

chucklever:
Force-pushed with two changes to patch 2, prompted by the bot review:

- The wait for sendmail to exit is bounded at 60 seconds, the same bound lettre puts on an SMTP session. The outbox is drained one message at a time, so an MTA that never exited would hold every later message. lettre spawns the child with kill_on_drop, so the timeout also reaps it.
- A sendmail_path that is not an executable file is now refused at startup, next to the public_base_url check, instead of the worker logging once and returning while the daemon runs on with mail queued forever. The check requires a regular file with an execute bit, since a directory or a 0644 file exists but spawn-fails all the same.

The hyphen-prefixed recipient finding does not apply: lettre passes `--` before the recipients and the sender as the operand of `-f`.

Also updated the settings review guide, which still listed smtp.server and smtp.port as required. Sign-offs valid, clippy clean, 803 unit and 43 integration tests passing.

chucklever:
Rebased onto d0e8bfce601a and force-pushed. The code is unchanged apart from context: the NNTPS commits that merged this morning add two tests to the settings.rs test block next to the smtp transport tests, and Cargo.lock now carries the asn1-rs packages the TLS transport records alongside the async-std packages the sendmail-transport feature records.

Sign-offs valid, clippy clean, 808 unit and 43 integration tests passing.

PR #510, opened 2026-09-15
api: match forge placeholder IDs to Git ingestion IDs

zraza-xlnx:
PR/MR webhooks create fetching placeholders without "@sashiko.local" suffix, but Git ingestion uses the suffix when looking up the patchset. This mismatch creates a second row in Web UI instead of updating the placeholder.

Append @sashiko.local to the webhook placeholder ID in src/api.rs:1249, matching Git ingestion and the existing manual Git submission path.

References:

- Git ingestion generates the suffixed ID
https://github.com/sashiko-dev/sashiko/blob/ea697a9190eeabff72b4049ddb4c3b106a6e6c3b/src/main.rs#L2280-L2288
- Ingestion expects the ID to match the placeholder
https://github.com/sashiko-dev/sashiko/blob/ea697a9190eeabff72b4049ddb4c3b106a6e6c3b/src/main.rs#L1970-L1972
- Manual Git submissions already use the suffix
https://github.com/sashiko-dev/sashiko/blob/ea697a9190eeabff72b4049ddb4c3b106a6e6c3b/src/api.rs#L464-L469


zraza-xlnx:
Fixed in other commits, https://github.com/sashiko-dev/sashiko/blob/68085717b3136625989d55d73c5318eb5bc5c1f7/src/api.rs#L1985

Thanks!

PR #513, opened 2026-09-16
baseline: route iwl-net and iwl-next series to dev-queue

pkitszel:
Intel Wired LAN patches are tagged iwl-net or iwl-next in the subject
and are queued on the dev-queue branch of Tony Nguyen's queue trees,
never on those trees' default branches. The generic subsystem heuristic
had no way to know this and picked the default branch, so the series
usually failed to apply.

Detect the tag in the subject (case insensitive) and emit a single
candidate pointing at dev-queue. The repository URL is not hardcoded:
it is taken from the MAINTAINERS T: trees that the touched files voted
for, with iwl-net mapping to net-queue and iwl-next to next-queue.
When no matching tree is present the normal heuristic still applies.

Assisted-by: Claude Opus 5
Signed-off-by: Przemek Kitszel <przemyslaw.kitszel@intel.com>


pkitszel:
@rgushchin who could review/merge this PR?

rgushchin:
Merged, thanks!

PR #514, opened 2026-09-17
cli: show the worker's progress during a local review, and its log tail when it fails

ftabba:
## Summary

`sashiko-cli local` spawns `sashiko worker` and reads its stderr only to drive the phase display, so from "AI review in progress" until the result nothing appears, and a review that has hung looks the same as one that's slow (a single-patch run here took 45 minutes today). When the worker fails before producing a result, its cause is discarded with the rest of the stream.

Three commits:

1. Keep a bounded tail (100 lines) of the worker's stderr and replay it when the worker exits with no result. This is #233, carried over unchanged.
2. Give the worker a progress callback that writes one plain `progress:` line per stage event (the planned stages, each stage's start and finish, its turn count against the budget, a retry, and the patch finishing), and have the CLI print them as they arrive under the phase header. Turn ticks overwrite one line when stderr is a terminal, are dropped when it isn't, and stay out of the tail either way.
3. Replay the tail on the parse-failure path as well, which the review comment on #233 raised: a worker that wrote something other than a result is the case where the tail helps most.

This supersedes #233, which I'll close. On its own the tail was plumbing that only helped on failure; the progress lines are what make the same stream useful on a successful run, and the two together turn a silent 45 minutes into this (stderr captured to a file, so no turn ticks; the seven analysis stages run concurrently, so their start and finish lines interleave):

```
[3/4] AI review in progress...
      reviewing 1 patch
      patch 1: pre-screen
      patch 1: planning
      patch 1: stages goal, implementation, execution-flow, resources, locking, security, hardware, deduplication, conflict-resolution, verification, report
      patch 1: goal started
      patch 1: implementation started
      patch 1: execution-flow started
      patch 1: resources started
      patch 1: locking started
      patch 1: security started
      patch 1: hardware started
      patch 1: resources finished
      patch 1: security finished
      patch 1: locking finished
      patch 1: goal finished
      patch 1: hardware finished
      patch 1: implementation finished
      patch 1: execution-flow finished
      patch 1: deduplication started
      patch 1: deduplication finished
      patch 1: conflict-resolution started
      patch 1: conflict-resolution finished
      patch 1: verification started
      patch 1: verification finished
      patch 1: report started
      patch 1: report finished
[4/4] Review complete.
      patch 1: done
```

The last line lands after the phase-4 marker because the worker logs "AI review completed", which the CLI keys the marker on, before it emits the patch's finished event. I've left that as is; if you'd rather the marker waited for the stream to end, that's a small change in the CLI loop.

## Testing

`make check-pr` passes except yamllint, which isn't installed here; no YAML is touched. One new unit test pins the turn and stage line formats and which events stay silent. End to end, one real review of a single commit produced the output above, with stderr going to a file, so the on-terminal tick display (the in-place overwrite) has not been seen running. The no-output and unparseable-output failure paths have no test, as #233 didn't, and I haven't exercised them end to end in this round.


kees:
Any reason you're using sashiko-cli instead of `sashiko review`? There is extensive progress reporting there already (and I've been cleaning it up recently like with PR #501 )

ftabba:
> Any reason you're using sashiko-cli instead of `sashiko review`? There is extensive progress reporting there already (and I've been cleaning it up recently like with PR #501 )

Simply because I didn't know about it. I've been using sashiko-cli since forever. Checking this out now, and I agree it fits my needs better. Thanks for pointing it out.

PR #519, opened 2026-09-17
Various fixes to minor issues in sashiko

helq:
While developing #518, my agent found multiple bugs in related portions of the code. They all sound sensible fixes but some might be unnecessary. I include the whole bunch and if you want me to drop any fixes in particular I can do that :)


rgushchin:
forward AI provider API keys to review child process - this is a very bad idea, which increases the chances that the API key can be stolen. Why would the process need it?

PR #520, opened 2026-09-18
ai: honour response_cache for local reviews

kees:
AiSettings carries response_cache, but only the daemon ever consulted it. Using "sashiko review" builds its provider with create_provider_from_ai, which is uncached, so setting response_cache did nothing on that path and gave no sign that it had not taken effect.

The problem was that the constructor derived the cache location from settings.database.url, and a local review holds AiSettings alone with no database beside it. Provide a cache location for local review or when the database location is a schema (i.e. not a path name) in $XDG_DATA_HOME/sashiko/.

A daemon-spawned worker is left uncached, on the same test the concurrency and retry decorators use: it reaches the model through a stdio provider, and the daemon holds a cache on the far side of it. Caching there as well would give one run two of them, filled from the same responses.

The helper resolving the data directory moves to utils, since the response cache and the prompt bundle both want it and neither belongs to the other.

Assisted-by: LLM
Co-developed-by: Kees Cook <kees@kernel.org>


kees:
This is a rebase and rework of PR #493 so that it avoids double caching and avoids schema paths.

kees:
Fixed! And now I have found `--project sashiko` for doing local reviews with the dedicated sashiko-specific prompts! :)

rgushchin:
@kees also sashiko.sashiko.dev

shardulsdk-mpiric:
@kees heads-up: opened #568 . The response cache misses for every stage with tools, because the tool list goes out in random order. A fix series with tests is linked there.

PR #524, opened 2026-09-18
review: ask the dirty-tree question on stderr, and only for tracked changes

ftabba:
## Summary

Following Kees's comment on #514 I switched my local runs to `sashiko review`, and the first run hung. The tree had build logs in it, so the dirty-tree check fired, and the `Do you want to proceed? [y/N]` prompt went to stdout, which under `--format json` was redirected to a file: nothing on the terminal, the prompt as the file's only content, and Enter aborted the run.

Two commits:

1. Write the prompt on stderr, where the warning it follows already goes. stdout is the review's output.
2. Leave untracked files out of the dirty-tree check (`git status --porcelain --untracked-files=no`), with a test. A kernel tree carries build logs, a `.config` and patch directories that aren't changes to anything the review reads, so the warning fired on every run. If you'd rather keep warning on untracked files, this commit can be dropped; the first stands on its own.

`make check-pr` passes on the branch (770 lib tests).


ftabba:
@kees Thanks for the suggestion. I found this after the switch.

PR #526, opened 2026-09-18
local_review: attribute each finding to the patch its locations name

ftabba:
## Summary

Every finding's `patch_index` is the index of the review pass that produced it, not the patch that contains it. A pass isn't confined to its own patch's code: on the 49-patch GICv5 IRS v5 series [1] the pass for patch 5, which touches only `arch/arm64/include/asm/sysreg.h`, returned four findings in `arch/arm64/kvm/vgic/*.c`, and one function drew nine different patch labels. Scored against the files each finding cites, the label was right for 6 of 24 inline blocks. Anything that joins findings to the series by patch number (dedup, a per-patch reply) then mis-attributes.

Two commits. The first: keep the pass in a new `review_pass_index` field, and make `patch_index` and `patch_subject` the one patch whose files the finding's `locations` name, resolved against the per-patch file lists the worker already gathers for its tool scope, now once for the series before the fan-out. A finding with no locations, in files no patch touches, or in files several patches touch keeps the pass index.

This is the code-only half. File names can't settle a finding in a file several patches touch; the model's own `commit <sha>` header in the inline review was right for 23 of the 31 headers on the same run. The follow-up I'll send separately, so it can be benchmarked on its own, has the verification stage return that commit as a structured field and uses it here when present.

The second, from sashiko's own review of earlier heads: the aggregation loop indexes each element of the findings and concerns arrays and nothing validated them as objects (`validate_concerns_output` was a no-op, the verification stage had no validator), so a string or null from the model panicked the worker. Every stage that returns one of these arrays, in the Linux and the Sashiko workflow alike, now rejects such an element by index and the engine retries with feedback (the Sashiko workflow shares the Linux validators instead of its no-op copy); the aggregation keeps a guard for a result that arrived without the validators and logs what it drops rather than dropping it silently. Pre-existing concerns also carry `review_pass_index`, from the same review.

`make check-pr` passes on the branch.

[1] https://lore.kernel.org/all/20260807111159.429128-1-sascha.bischoff@arm.com/


ftabba:
Pushed a new head: the first commit now stamps review_pass_index on pre-existing concerns too (sashiko's MEDIUM), and a second commit drops a findings or concerns element that is not a JSON object instead of panicking in IndexMut (its pre-existing HIGH).

ftabba:
New head for the second commit: instead of dropping a non-object array element silently, the analysis, dedup and verification stages now reject it by index so the engine retries with feedback (the validator hook was a no-op for these arrays); the aggregation keeps a guard for results that arrive without the validators and logs each drop.

ftabba:
New head for the second commit, after sashiko's review of the previous one: the validators now cover the conflict-resolution stage and the whole Sashiko workflow, which shares the Linux validators instead of its no-op copy. On its other point, a missing findings or concerns key deserialising as empty under serde(default): that's upstream's existing choice, and turning it into a retry trades a possible false-clean for a failed review once the retries run out, so I've left it as is; say if you'd rather it retried.

ftabba:
On "Sashiko review — v4", none changed:

patch_index in the daemon: it reviews one patch per worker, stores each finding against that review's id and replies on that patch's mail; the description says the change is visible on the local paths.

patch_shas in daemon mode: --review-commit puts one patch in it, so the derivation falls back to the pass index there, as before this PR.

Dropping a non-object element with a warning: the stage validators are the check; the aggregation guard only runs for a result that arrived without them, and failing the review there would lose the valid findings over one element.

Missing array keys under serde(default): the existing deserialisation choice; a retry there trades a possible false-clean for a failed review.

PR #527, opened 2026-09-18
workflow: let verification name the patch that introduced a finding

ftabba:
## Summary

Follow-up to #526, stacked on it (its two commits come first here). File names can't attribute a finding in a file several patches touch, and a series that builds one file up patch by patch is that case throughout; on the GICv5 IRS v5 run [1] one function drew nine patch labels. The review's own inline prose already carried the right commit header for 23 of the 31 headers there, so this requests it as data at the stage that has the series in front of it.

One commit: each finding the verification stage returns gains `introduced_in_patch`, the series position of the patch whose change first made the problem present, chosen from the series list, or null when the code came from before the series or the stage cannot resolve it. The series block now lists the preceding patches as well as the subsequent ones, so every position is nameable (and the last patch of a series gets the block too). The validator accepts an integer or null. The aggregation takes the named patch only where the cited files allow it (a series patch and, when any patch touches those files, one of them), so it settles what #526's file-name derivation leaves open and can't override it; a null, a contradicted value or an older prompt bundle behaves as #526 does.

This changes the verification prompt and its series block, so it's the half to benchmark; #526 is the code-only half. `make check-pr` passes on the branch.

[1] https://lore.kernel.org/all/20260807111159.429128-1-sascha.bischoff@arm.com/


ftabba:
On "Sashiko review — v1", benchmark data: benchmark_tiny.json, one run each, gemini-3.1-pro-high, temperature 1.0. Before this commit (the #526 head): 7 detected, 0 partial, 2 missed, 113 concerns, 49 findings. After: 7, 0, 2, 105 concerns, 40 findings. One entry swapped each way between detected and missed, the run-to-run variance at nine entries. introduced_in_patch was filled on all 16 verification findings, 13 as the reviewed patch and 3 null, none the validator had to reject. The suite is single-patch, so the cross-patch value isn't exercised here; the full benchmark is yours to run.

Issue #530, opened 2026-09-20
Other gregkh-maintained trees likely have the same wrong-HEAD bug as staging (#268)


HadiAbu:
Follow-up to #268 / #529 (staging-only fix).

`gregkh/staging.git`'s default `HEAD` symref points at `main`, which
tracks what has already merged upstream rather than the branch new
patches apply to (`staging-testing`). #529 fixes this specifically
for the STAGING SUBSYSTEM tree.

The same pattern holds for every other tree Greg Kroah-Hartman
maintains — checked directly against kernel.org:

```
$ for repo in usb char-misc driver-core tty; do
    git ls-remote --symref "https://git.kernel.org/pub/scm/linux/kernel/git/gregkh/$repo.git" HEAD | head -1
  done
ref: refs/heads/main	HEAD   (usb)
ref: refs/heads/main	HEAD   (char-misc)
ref: refs/heads/main	HEAD   (driver-core)
ref: refs/heads/main	HEAD   (tty)
```

All four have `HEAD` pointing at `main` rather than their
`<name>-testing` / `<name>-next` branches, the same convention as
`staging`. Patches for USB, char-misc, driver-core, and tty
subsystems likely get silently applied against the wrong revision
the same way #268 described for staging, just unreported so far.

Worth generalizing the fix in `resolve_subsystem_heuristic`
(`src/baseline.rs`) to detect any `gregkh/<name>` tree and try
`<name>-testing` / `<name>-next` instead of hardcoding one
subsystem at a time, rather than adding a near-identical special
case per tree as each gets reported.

PR #531, opened 2026-09-20
baseline: use staging-testing for the STAGING SUBSYSTEM tree

### LINKED ISSUE #268: Change base repo for linux-staging
`linux-staging` patches are expected to be applied against Greg Kroah-Hartman's `staging-testing` branch:

https://git.kernel.org/pub/scm/linux/kernel/git/gregkh/staging.git

As of 2026-06-11, Sashiko reports that it successfully applied and reviewed the following patch:

https://sashiko.dev/#/patchset/20260611103748.29584-1-ixebec777%40gmail.com

However, the patch does not apply cleanly to the current `staging-testing` branch. At the time of writing, `staging-testing` is at commit:

`7cb1c5b32a2bfde961fff8d5204526b609bcb30a`

This suggests that Sashiko may be validating `linux-staging` submissions against a different repository, branch, or revision than the one expected by the subsystem.

HadiAbu:
## Summary

Fixes #268. Sashiko was applying `linux-staging` patches against the
wrong revision of the tree.

## Root cause

`gregkh/staging.git`'s default `HEAD` symref points at `main`, which
tracks what has already been merged upstream, not the branch new
patches are expected to apply to. Verified directly against the live
tree:

```
$ git ls-remote --symref https://git.kernel.org/pub/scm/linux/kernel/git/gregkh/staging.git HEAD
ref: refs/heads/main	HEAD
df2908090cda368b01ff43709f51890076c56157	HEAD

$ git ls-remote --heads https://git.kernel.org/pub/scm/linux/kernel/git/gregkh/staging.git
...
df2908090cda368b01ff43709f51890076c56157	refs/heads/main
df2908090cda368b01ff43709f51890076c56157	refs/heads/staging-linus
8444548bd905f22093729065408284a6b46f7eee	refs/heads/staging-next
9dbdac024d4d9f3fc234399cb8c1f1cc6d2bcb2d	refs/heads/staging-octeon
8444548bd905f22093729065408284a6b46f7eee	refs/heads/staging-testing
```

`main` and `staging-linus` sit at the same commit; `staging-testing`,
where incoming `drivers/staging/` patches actually land, is ahead of
both. The MAINTAINERS `T:` line for STAGING SUBSYSTEM carries no
branch, so `resolve_subsystem_heuristic` in `src/baseline.rs` fell
back to `<remote>/HEAD` and applied series against the stale branch.

## Fix

Added a special case in `resolve_subsystem_heuristic` mirroring the
existing linux-mm handling in the same function (which exists for an
identical reason — `akpm/mm.git`'s HEAD isn't the right branch
either): when the matched tree is `gregkh/staging` or the subsystem
name is STAGING SUBSYSTEM, prioritize `staging-testing` then
`staging-next` over the generic HEAD fallback.

## Scope note

I checked, and this same HEAD-points-at-`main` pattern affects every
other GregKH-maintained tree (`usb.git`, `char-misc.git`,
`driver-core.git`, `tty.git`), not just staging — they likely have
the identical silent-misapply bug, just unreported so far. Keeping
this PR scoped to the one reported subsystem; opened #530 to track
generalizing the fix to the rest of GregKH's trees.

## Testing

- Verified the actual branch divergence against the live
  `gregkh/staging.git` remote (see above) rather than assuming from
  memory.
- Added `test_resolve_staging` in `src/baseline.rs`, mirroring the
  existing `test_resolve_linux_mm` test.
- `cargo test --all-features`: 841 tests passed (777 lib + 41 main +
  4 CLI + 19 integration, 16 ignored).
- `cargo clippy --all-targets --all-features -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.


HadiAbu:
Fixed both in 2c93e6e:

- **HIGH**: `resolve_subsystem_heuristic` now keys the Staging/MM special cases off the subsystem tied to the actual winning tree (`tree_subsystems`, tracked per-tree) instead of whichever file matched first while scanning. Added a regression test (`test_resolve_staging_does_not_hijack_other_subsystem`) covering a series where a Staging file is iterated first but a different subsystem's files are the majority.
- **MEDIUM**: `Signed-off-by` trailer now uses the full name, `Hadi Abu <hadi.abuhamed@gmail.com>`.

PR #533, opened 2026-09-20
feature: stages in report output

kees:
I wanted to be able to see what stages are contributing to a finding, so plumb the stages through the results, the database, and the reports. Now we can see the stages along with the severity, e.g.:

```
- [CRITICAL] (Concurrency & Async, DB & Persistence) In src/worker/sync.rs
  (GitSyncWorker::sync_all_remotes), holding the synchronous
  std::sync::MutexGuard across the async fetch_remote() call can deadlock
  Tokio worker threads when multiple remotes sync concurrently.
 
- [HIGH] (Security Audit) In src/worker/sync.rs
  (GitSyncWorker::sync_all_remotes), unredacted git fetch stderr is logged
  via warn! and error! when remote URLs fail, leaking embedded
  authentication tokens into application logs.
```

A couple sashiko-review prompts needed adjustment: it kept throwing a false positive on json insert, and it had some conflicting ideas about how to do safe database migrations.


kees:
> ### Sashiko review
> #### Commit 1/7 — `67eb6661` prompts: teach the self-review what serde_json indexing does
> * [MEDIUM] In prompts/sashiko/false-positive-guide.md, the guide explicitly
>   commands the verification stage to read error-handling.md. Since this file
>   is not in the verification context, the LLM will attempt to fetch it via
>   the read_prompt tool. However, the cross-reference omits the required
>   patterns/ directory prefix. Calling read_prompt with just the filename
>   will fail with a file not found error, pointlessly burning a tool turn,
>   wasting tokens, and risking the LLM hallucinating the file contents. The
>   instruction should provide the full relative path or avoid the command
>   entirely since the rules are duplicated inline.

I decided to just drop the mention instead; it was technically redundant anyway. Fix pushed.

> #### Commit 3/7 — `0c83de1a` review: carry a concern's stage through consolidation for sashiko
> * [HIGH] In src/workflows/sashiko_patch_review.rs
>   (STAGE_DEDUPLICATION_INSTRUCTION, STAGE_VERIFICATION_INSTRUCTION, and
>   deduplication_stage), modifying core instructions for the deduplication,
>   conflict resolution, and verification stages risks silently degrading AI
>   review quality. Under Sashiko's mandatory review guidelines, any change to
>   stage instructions, deduplication logic, or global prompts must be backed
>   by benchmark data. A review of this patch confirms no benchmark evaluation
>   data is included.

I don't have a way to benchmark this (only the Linux workflow has benchmarks as far as I can see), but FWIW, it has had no problem finding errors in my patches after these changes. :)

kees:
> ### Sashiko review — v2
> #### Commit 2/7 — `436968b7` review: carry a concern's stage through consolidation
> * [HIGH] In src/workflows/linux_patch_review.rs (deduplication_stage,
>   conflict_resolution_stage, and verification_stage), calling
>   stages_that_raised with an array of Vec references causes an E0308
>   compilation error. An array of thin pointers to Vecs does not implicitly
>   coerce to the expected slice of fat pointers, which completely breaks
>   the build.

Uh, what? I'll double check, but I think this is a hallucination. e.g. I don't have build failures.

> #### Commit 3/7 — `ddcd8956` review: carry a concern's stage through consolidation for sashiko
> * [HIGH] In src/workflows/sashiko_patch_review.rs
>   (STAGE_DEDUPLICATION_INSTRUCTION, STAGE_CONFLICT_RESOLUTION_INSTRUCTION,
>   CONCERN_JSON_SCHEMA_EXAMPLE), the commit modifies core instructions and
>   JSON schema expectations to enforce new provenance tracking rules. Changing
>   core instructions directly influences the LLM's capacity to consolidate
>   bugs and format findings correctly. Under Sashiko's global review
>   guidelines, any change that can meaningfully affect the overall quality of
>   AI reviews across the board MUST be backed up by benchmark evaluation data
>   in benchmarks/. No benchmark data was added or updated to validate these
>   prompt alterations.

Same request for benchmarks. I'm open to suggestions for this being meaningful for sashiko. The Linux benchmark details are in patch 1.

> * [HIGH] In src/workflows/sashiko_patch_review.rs (verification_stage) and
>   src/workflows/linux_patch_review.rs (verification_stage,
>   test_assigning_provenance_inserts_a_key_that_was_not_there), a new concern
>   object is instantiated via the json! macro without a stages key. The code
>   then attempts to insert the key using concern["stages"] = stages.clone();.
>   However, serde_json::Value::IndexMut for &str on an Object explicitly
>   panics if the key does not already exist. Contrary to the author's
>   assumption, this will unconditionally panic and crash the Tokio worker
>   process anytime an LLM correctly reports a pre-existing finding with a
>   stages field. To fix this, concern.as_object_mut().unwrap().insert(
>   "stages".to_string(), stages.clone()); must be used.

The hallucination I got the whole time: and it's objectively false. The test code explicitly runs the code this feedback says will panic. :)

rgushchin:
I'm not sure about it: I totally get the value for development of Sashiko (this part I'm happy to merge), but not sure I want to expose it to all kernel developers. It's generally not actionable for most users and might confuse people when looks wrong. How about making it an option for local reviews?

kees:
That's fine by me, yeah. I'm a bit less sure how to approach it. It requires changes to the stage prompts (to pass the stage name around) and making that optional looks messy. Do we want to just limit the reporting? i.e. retain the stage details and keep the db schema update, but make the reporting template part optional?

rgushchin:
Yeah, I guess db schema update is fine. It's actually useful information for sashiko development (e.g. relative usefulness of various stages). Btw, I pushed a new sashiko for sashiko, should be less noisy hopefully.

PR #535, opened 2026-09-20
cache: print token counts without a thousands separator

kees:
fmt_thousands() groups with a period, so a log line reads "3.923 tokens saved (total previous session: 335.333)". That is deliberate: 3ecb53a says "dot separators", which is the convention in its author's locale. It does not survive contact with the rest of the tree.

335.333 is a valid number under either convention and is additionally confusing since there's no such thing as decimal token counts. Logging output should not vary with the deployment's locale either, which is the reason Rust's standard library offers no grouping at all. This repo writes American English, which groups with a comma if it groups. And an unseparated integer sorts and greps correctly as text, which is worth more in a log line than readability at a glance.

Consistency also argues for this change. Every other token count in the tree prints raw: the review's own footer in main.rs and sashiko-cli.rs, "Tokens: {} in / {} out / {} cached", and the benchmark's "Avg Tokens In". These cache lines were the only place that grouped, so dropping the grouping makes the tree agree with itself rather than carry a second convention in one corner of it.

So this is the revert of 3ecb53a, plus the call sites 25fe048 added after it. Those group the hit counts as well, which are not token counts at all and reach three digits only in a long run.

If output a person reads rather than greps ever wants grouping, a web UI say, num-format's to_formatted_string() with a locale taken from the request is the tool for it, rather than one separator chosen once for the whole process.


kees:
cc @acmel

acmel:
On Mon, Sep 21, 2026 at 01:41:02AM -0700, Roman Gushchin wrote:
>    Merged [1]#535 into main.

Agreed, brown paper bag on my part, should've used locales, thanks a lot
for fixing it.

- Arnaldo

PR #557, opened 2026-09-23
review: show the series cover letter to the reviewer

matthiasgoergens:
This reworks the change along the lines @rgushchin suggested: the series' cover letter is now shown to the verification stage only, it is on unconditionally, and the `--cover-letter` option is gone. The cover letter is where authors usually explain the shape of a series, such as why one patch depends on another, and until now a review saw each patch's diff and the other patches' subject lines but never that explanation.

The cover letter is text that only its author vouches for, so it stays out of the shared system prompt. It is routed the way the commit message is: the stages that trace the code's mechanics (execution-flow, resources, locking and security) get the hunks without the commit message, so that its claims cannot bias them, and that choice is declared per stage in the stage table. The cover letter uses the same kind of declaration: `ConsolidationStage` gains `wants_cover_letter`, which is set only for verification. The analysis stages, deduplication and conflict resolution never see the cover letter, so every concern is raised without it. Only verification receives the raw cover letter; its model-generated findings, including severity and explanation, are subsequently consumed by the report stage. That is also where a misleading cover letter could try to argue a real finding away, so the block is written to fit the rule the stage already has: to discard a concern it needs concrete evidence in the code. The block says the text is the author's account, not instructions, that its claims are not evidence, and that a concern is not to be discarded because the cover letter says the problem is handled, intended or tested. The series validation rule, which already said not to trust promises in commit messages, now names the cover letter as well. The block sits in the verification stage's user prompt after the series context, and `&`, `<` and `>` in the letter are escaped, so that the text cannot close the block, open another, or imitate one of the prompt's own blocks such as `<pre_fetched_context>`; the block tells the model about the escaping. The Sashiko workflow gets the same block in its own verification stage.

There is no option. A cover letter reaches a review in one way only, through an optional `cover_letter` field of `ReviewInput`, the payload the worker reads, and it is set whenever the input carries one. In the daemon, the reviewer fills it from the ingested cover letter, as in the previous version: `Database::get_cover_letter_body` only returns the body of a message that announces part 0, has no reply prefix, comes from the series' author and is not blank, because `cover_letter_message_id` names the series rather than always a cover letter. A local review of a commit range fills it from the range itself when the range records one in either of two conventions. If the range ends in a two-parent merge commit, every other commit of the range is on its merged side, and the merge's tree is exactly what `git merge-tree --write-tree` gives for its parents without conflicts, the current heuristic treats the range as a no-ff-style series merge and takes the merge's message as its cover letter, as kernel trees record a series' or a pull's text in the merge. A merge that resolved a conflict or carries changes of its own stays in the review, so taking the merge out never hides a change. Clean-merge cover extraction requires Git 2.38 or newer; on older Git the merge is retained as a patch and no cover letter is extracted. The shape cannot prove provenance: a range from a topic branch's tip to a merge of mainline into it qualifies too, and the merge message is then shown as the cover letter. If the first or last commit of the range is a `b4 prep` cover commit (an empty commit with one parent, non-blank text above b4's `--- b4-submit-tracking ---` line, and b4's JSON tracking data with a `series` object below it), the text above that line is the cover letter. In both cases the commit that holds the letter is not reviewed as a patch. A range that also reaches first-parent commits behind the merge keeps its merge commit as a patch, and a single commit is always reviewed as it is. `sashiko-cli`'s local path had its own copy of the range handling and now calls the same `build_review_input_from_git`, so both local entry points resolve a range the same way. One difference remains: a range submitted to the daemon through `/api/submit`, and a forge pull request, gets a synthetic series identity and so no cover letter, while the same range reviewed locally may have one. Carrying the range's cover letter through the git fetcher would close that, and I have left it out of this PR to keep the change small.

Without a cover letter the only prompt change is a few words in the series validation rule of each workflow's verification stage, which now names the cover letter. The suites in `benchmarks/` contain only single commits, so they should barely move, if at all. The effect on review quality is unmeasured, and I expect the full benchmark will be needed before this can be merged. The meaningful test is to review multi-patch series with and without their cover letters and compare the findings, the false positives and the token cost.

The tests record every request a worker sends for a two-patch series run through all stages, and check that the cover letter appears only in the verification stage's user prompt and in no other request, and that without a cover letter nothing is rendered anywhere. Other tests cover the block's rendering and the escaping of injected opening and closing tags and of a fake `<pre_fetched_context>` block, the stage tables of both workflows, the payload with and without the field, a payload in the old format, the database lookup (a real cover letter, a renamed series, a single patch, a reply, another author's part 0, a blank letter, a missing patchset), and local ranges ending in a clean series merge, in a merge that resolved a conflict, in a merge with a change of its own, in mainline merged into a topic branch, reaching past the merge base, starting or ending with a b4 cover commit, an empty commit without the marker, a non-empty commit carrying it, marker commits with a blank letter, no tracking data, unparseable tracking data or JSON that is not b4's, a range made only of a cover commit, and a range with no cover letter. `make check-pr RANGE=origin/main..HEAD` passed (sign-offs on all three commits, clippy with `-D warnings`, `cargo fmt --check`, `yamllint` and `cargo test --all-features`); its test runs reported 16 ignored tests. `make check-integration` was not run.


rgushchin:
I think we should be careful with this. I'm not saying it's always a bad idea, but I'm concerned about opening a possible prompt injection attack vector. Even the commit log is hidden from some stages exactly because of this reason. I need to think more about it, but maybe we can include it into the verification stage only or maybe even a dedicated stage.
Also I'd not add this as an option: it creates different way to run sashiko and get different results. I'd rather avoid it and turn it on unconditionally but in a way we can trust.

matthiasgoergens:
Thanks, that's a fair concern. Having the cover letter in the review is a nice-to-have, and the prompt-injection risk may well outweigh it.

In case it helps to see it, I've pushed a rework along your lines. There is no option any more: the cover letter is used whenever the review input has one, and only the verification stage receives it, in its user prompt rather than the shared system prompt, so the stages that read the hunks on their own don't see it, just as they don't see the commit message. The prompt tells the model the letter is the author's claims, not instructions, and that a concern can only be dismissed on evidence in the code. For local reviews the cover letter comes from the range itself: a clean no-ff merge commit at the tip, or a `b4 prep` cover commit.

If you'd rather contain it further, a dedicated stage would work too, and I'm equally happy to drop this; #556 was the part that fixed an existing bug, and it's in.

